PR #13, opened 2015-06-29
HWKINVENT-78 Remove version from ResourceType

ppalaga:
This is a preparation for the upcoming Inventory 0.1.0.Alpha3


ppalaga:

PR #19, opened 2015-07-02
Mazz/counter metrics

jmazzitelli:


hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/storage/MetricsOnlyMetricDataPayloadBuilder.java
```diff
@@ -33,18 +35,38 @@
...
+    public void addDataPoint(String key, long timestamp, double value, MetricType metricType) {
+        Map<String, List<Map<String, Number>>> map;
+
+        switch (metricType) {
+            case GAUGE: {
+                map = allGauges;
+                break;
+            }
+            case COUNTER: {
+                map = allGauges; // SHOULD BE allCounters!!!
```
jmazzitelli: when https://issues.jboss.org/browse/HWKMETRICS-166 is fixed, we can change this line to be "map = allCounters;"
and delete the log code below.

PR #23, opened 2015-07-17
agent 0.3.1-SNAPSHOT + inventory 0.2.0-SNAPSHOT

jkremser:
On the hawkular/hawkular side this branch (https://github.com/Jiri-Kremser/hawkular/tree/agent-0.3.1-with-inventory-0.2.0) is the right one to work with this PR 

**don't merge yet**

### DIFF
--- a/hawkular-wildfly-monitor/src/main/resources/module/main/module.xml
+++ b/hawkular-wildfly-monitor/src/main/resources/module/main/module.xml
@@ -23,6 +23,7 @@
     <resource-root path="hawkular-bus-rest-client.jar"/> <!-- DELETE THIS ONCE WE NO LONGER SEND TWO MESSAGES -->
     <resource-root path="hawkular-dmr-client.jar"/>
     <resource-root path="hawkular-feed-comm-api.jar"/>
+    <resource-root path="hawkular-inventory-json-helper.jar"/>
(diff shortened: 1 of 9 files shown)


hawkular-wildfly-monitor/src/main/resources/module/main/module.xml
```diff
@@ -23,6 +23,7 @@
     <resource-root path="hawkular-bus-rest-client.jar"/> <!-- DELETE THIS ONCE WE NO LONGER SEND TWO MESSAGES -->
     <resource-root path="hawkular-dmr-client.jar"/>
     <resource-root path="hawkular-feed-comm-api.jar"/>
+    <resource-root path="hawkular-inventory-json-helper.jar"/>
     <resource-root path="hawkular-monitor.jar"/>
     <resource-root path="inventory-api.jar"/>
```
jmazzitelli, line 28: based on the changes to the pom.xml, I assume this "inventory-api" was renamed so this should be removed from module.xml

jkremser, line 28: nope, `hawkular-inventory-json-helper` has the `inventory-api` as a transitive dependency, so there is no need to have both in `pom.xml`, but I guess, in `module.xml` all the jars need to be explicitly mentioned.

PR #27, opened 2015-07-29
Debug output when searching for operations to perform

ppalaga:


ppalaga:
@jmazzitelli this helped me to find out that I need tho send `Redeploy` rather than `redeploy`.

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/feedcomm/ExecuteOperationCommand.java
```diff
@@ -89,9 +89,15 @@ private ExecuteOperationResponse executeOperationDMR(ExecuteOperationRequest req
         Address opAddress = null;
         String actualOperationName = null;
 
+        String requestedOpName = request.getOperationName();
         Collection<DMROperation> ops = resource.getResourceType().getOperations();
+        MsgLogger.LOG.tracef("Searching for operation [%s] among %d operations available for resource [%s].",
+                requestedOpName, ops.size(), resource.getID());
```
jmazzitelli: rather than logging each time in the loop, why not just log all the operations here in this one trace? that way, one trace log message provides you with all the information you need:

 MsgLogger.LOG.tracef("Searching for operation [%s] among operations [%s] for resource [%s].",
                requestedOpName, ops, resource.getID());

You can then eliminate the changed code below. Now you will have all the information in a single log message (with the added benefit of not having to make a method call ops.size() when trace isn't enabled)

ppalaga: +1, you are right, it can better be done in a single message.

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/feedcomm/ExecuteOperationCommand.java
```diff
@@ -89,9 +89,11 @@ private ExecuteOperationResponse executeOperationDMR(ExecuteOperationRequest req
         Address opAddress = null;
         String actualOperationName = null;
 
+        String requestedOpName = request.getOperationName();
         Collection<DMROperation> ops = resource.getResourceType().getOperations();
+        MsgLogger.LOG.tracef("Checking if operation [%s] is supported by resource [%s].", requestedOpName, resource);
```
jmazzitelli: You aren't logging the names of all valid operations though. From previous comment:

```
MsgLogger.LOG.tracef("Searching for operation [%s] among operations [%s] for resource [%s].", requestedOpName, ops, resource.getID());
```

notice "among operations [%s]" - and notice passing the "ops" collection.

This way, your log message will show you three things:

1) the operation that is being requested
2) the resource whose operation is being requested
3) the list of valid operations for that resource

So the log message would look something like:

   Searching for operation [deploy] among operations [{Deploy,Undeploy,Redeploy,Remove}] for resource [hostn~Local~/]."

From that one message, you'll be able to see "ahh.. I passed in a bad operation name - "deploy" is all lowercase, but the operation names are captizalized"

I don't know what the toString will look like for DMROperations, its probably something like "DMROperation[Deploy]" - but we'll get the idea.

ppalaga:
Sorry, you were right again. I thought for some reason that `resource.toString()` shows also the supported operations, which it does not. 01467f2 is completely adopting of your proposal.

PR #29, opened 2015-08-04
HAWKULAR-501 Let agent collect and store IP addresses of application servers

ppalaga:


ppalaga:
@jmazzitelli this the raw idea that I am going to test. I am showing this to you early, for the case that you'd like to comment.

ppalaga:
@jmazzitelli I have tested this for the collocated case and it delivers quite plausible IP addresses.

ppalaga:
The version in f855094 attempts to combine the two sources of IP address of an AS: 
(1) DMR of the AS and
(2) InetAddress.getByName(host) where host is the host the agent uses to query the AS'es DMR.

This works well, see https://github.com/hawkular/hawkular-agent/commit/f8550946747111d0e923141bc84445f7be8676a5#diff-6bc5e02300bbc2e87957a09fae5dc676R897 and https://github.com/hawkular/hawkular-agent/commit/f8550946747111d0e923141bc84445f7be8676a5#diff-6bc5e02300bbc2e87957a09fae5dc676R902 
The only case I am not sure what to present in the UI is when the AS is bound to the loopback interface: https://github.com/hawkular/hawkular-agent/commit/f8550946747111d0e923141bc84445f7be8676a5#diff-6bc5e02300bbc2e87957a09fae5dc676R889 ATM, I am sending 127.0.0.1 to the UI, which I personally find useless. As I said before, the IP address in the UI should help the user to know where to go to fix a problem he was just alerted about. 127.0.0.1 does not help in that respect at all. We could perhaps display `InetAddress.getByName(host)` also in this case, but the addresses should be somehow decorated (greyed out?) to indicate that the AS is not bound to them. WDYT?

pilhuhn:
Loopback is not useless. If the AS _public_ interface is explicitly bound to localhost, then the user has a reason.
It may be that there are IP table rules in place or that the AS is running inside a Docker container with its forwarding, but I believe that this is out of scope for now and in a later stage needs to be addressed by additional graph walking in inventory (following some 'runs-inside' relation or so).

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
```diff
@@ -819,4 +832,85 @@ private void connectToFeedCommChannel() throws Exception {
...
+        }
+
+        @Override
+        public void storeResourceType(ResourceType<?, ?, ?, ?> resourceType) {
+            delegate.storeResourceType(resourceType);
+        }
+
+        @Override
+        public void storeResource(Resource<?, ?, ?, ?, ?> resource) {
+            final String WF_TYPE_NAME = "WildFly Server";
```
pilhuhn: Isn't that something for a agent-wide constant, that is defined once?
Also suppose that in the future layered stuff would be deployed on top of WF - e.g. Hawkular, where the type could be "Hawkular Server <extends> WildFly Server", how would that work together?

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
```diff
@@ -819,4 +832,85 @@ private void connectToFeedCommChannel() throws Exception {
...
+            final String IP_ADDRESSES_PROPERTY_NAME = "IP Addresses";
+            String typeName = resource.getResourceType().getName().getNameString();
+            if (WF_TYPE_NAME.equals(typeName) && resource instanceof DMRResource) {
```
pilhuhn: I guess in the case of extension, that IP address would still come from the base type, so we should be good here?

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
```diff
@@ -819,4 +832,85 @@ private void connectToFeedCommChannel() throws Exception {
...
+                DMRResource dmrResource = (DMRResource) resource;
+                /* modify the config so that only the bound addresses are added */
+                DMRResourceConfigurationPropertyInstance adrProp = null;
+                for (DMRResourceConfigurationPropertyInstance prop : dmrResource.getConfigurationProperties()) {
+                    if (IP_ADDRESSES_PROPERTY_NAME.equals(prop.getName().getNameString())) {
```
pilhuhn, line 875: How is that checking for a/the public interface?
Suppose you have a dual-homed AS with the public (http) interface on 192.168.1.2 and the privat (management) interface on 10.1.2.3 then we want 192.168.1.2 returned and not the 10.\* address

ppalaga:
> Loopback is not useless. If the AS public interface is explicitly bound to localhost, then the user has a reason.

Yes, I agree that showing loopback may be useful for something. But it still does not help at all to know where to go to fix a problem the user was just alerted about. OK, let's ignore this for now.

So if we will show loopback addresses for loopback-only hosts, I will remove the non-loopback filter here https://github.com/ppalaga/hawkular-agent/commit/f8550946747111d0e923141bc84445f7be8676a5#diff-6bc5e02300bbc2e87957a09fae5dc676R898

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
ppalaga: Yes, it is tricky. `WF_TYPE_NAME` is supposed to be the same as the name of this `resource-type-dmr` : https://github.com/hawkular/hawkular/blob/4d60a3f1b846403734776317d81c2475b460616f/dist/src/main/resources/wildfly/patches/standalone.xsl#L850

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
ppalaga: > I guess in the case of extension, that IP address would still come from the base type, so we should be good here?

I understand all words but I have no idea what you are asking. What do you mean with "in the case of extension" and "base type"?

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
ppalaga, line 875: > How is that checking for a/the public interface?

The data in `DMRResourceConfigurationPropertyInstance prop` comes from this path in the DMR hierarchy: https://github.com/ppalaga/hawkular/commit/3ef5c839a9e7782ef7818563d05fcfd81ffe1428#diff-e2136f7d72c510834df4ad85535af167R861 which is the bound address of the http socket. (And it was you who said that we want http rather than management socket here.)

ppalaga:
BTW, I see `127.0.0.1` in the same class of usefulness as `0.0.0.0` - i.e. it tells something about the configuration of the server but does not help to identify the server. 

@pilhuhn Do we want to have that identification aspect in the AS list in the UI at all? I mean should I bother with replacing of `0.0.0.0` with the actual host addresses here https://github.com/ppalaga/hawkular-agent/commit/f8550946747111d0e923141bc84445f7be8676a5#diff-6bc5e02300bbc2e87957a09fae5dc676R897 ? Removing that part would make the fix for HAWKULAR-501 much simpler: actually adding the `resource-config-dmr` element in hawkular's standalone.xml would suffice: https://github.com/ppalaga/hawkular/commit/3ef5c839a9e7782ef7818563d05fcfd81ffe1428#diff-e2136f7d72c510834df4ad85535af167R862

ppalaga:
8aac54c makes Agent compatible with https://github.com/hawkular/hawkular/pull/396 and still replaces `0.0.0.0` with the list of resolved host addresses.

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
pilhuhn, line 875: On 7 Aug 2015, at 12:08, Peter Palaga wrote:

> which is the bound address of the http socket. (And it

Ok, I got confused by the "IP Addresses".
I was not aware that the binding now has the bound-address
available as a runtime-attribute. In the past we had to go
through a lot of hoops for this.

> was you who said that we want http rather than management socket here.)

Yes indeed.

## 

hwr@pilhuhn.de  m:0179/207 4919  b:http://pilhuhn.blogspot.com

pilhuhn:
On 7 Aug 2015, at 12:29, Peter Palaga wrote:

> BTW, I see `127.0.0.1` in the same class of usefulness as `0.0.0.0` - 
> i.e. it tells something about the configuration of the server but does 
> not help to identify the server.

Which in this case we try to get the server's ip address that matches
the canonical name of it (=the host).

ppalaga:
@jmazzitelli , @pilhuhn seems to agree with the present proposal. Could you please review and eventually merge?

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
```diff
@@ -512,11 +518,18 @@ private void startScheduler() throws Exception {
             try {
                 for (DMRInventoryManager im : this.dmrServerInventories.values()) {
                     resourceManager = im.getResourceManager();
+                    InventoryStorage invStorage = inventoryStorageProxy;
+                    try {
+                        invStorage = new ServerAddressResolver(im.getManagedServer(), inventoryStorageProxy);
+                    } catch (UnknownHostException e) {
+                        MsgLogger.LOG.error("Unable to get an IP address for managed server [" + im.getManagedServer()
+                                + "]", e);
```
jmazzitelli: trivial thing - but I have been using the "f" versions of the logger for things like this. (because I plan on doing a search-n-replace for all "errorf" and "warnf" calls to replace them eventually with org.jboss.logging generated methods in MsgLogger, so keep with this pattern). So this should be:

MsgLogger.LOG.errorf(e, "Unable to get an IP address for managed server [%s]", im.getManagedServer());

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
```diff
@@ -819,4 +832,74 @@ private void connectToFeedCommChannel() throws Exception {
...
+        public void storeResourceType(ResourceType<?, ?, ?, ?> resourceType) {
+            delegate.storeResourceType(resourceType);
+        }
+
+        @Override
+        public void storeResource(Resource<?, ?, ?, ?, ?> resource) {
+            final String WF_TYPE_NAME = "WildFly Server";
+            final String IP_ADDRESSES_PROPERTY_NAME = "Bound Address";
+            String typeName = resource.getResourceType().getName().getNameString();
+            if (WF_TYPE_NAME.equals(typeName) && resource instanceof DMRResource) {
```
jmazzitelli: What if you just checked for a null parent? That would tell you its the top root resource which is the one you want. No need for having it to be named something specific like "WildFly Server" -  for example:

if (resource instanceof DMRResource && resource.getParent() == null)

That way we know its the top root resource without having to know its actual name. I think that would work and would mean you can get rid of WF_TYPE_NAME completely.

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
ppalaga: +1. Yes, `*f()` is also my personal preference. I must have blindly copied it from somewhere. And now when looking at `ServerAddressResolver` again, the constructor does not need to throw anything anymore. Let me fix it.

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
ppalaga: OK, let me do that if you think the solution is going to be durable enough.

ppalaga:
@jmazzitelli 3e7ef22 incorporates both of your proposals.

PR #30, opened 2015-08-04
Even more of 'Hawkular-side' part of the HWKINVENT-107

ppalaga:


ppalaga:
@jmazzitelli a small fix of the previous commit

PR #31, opened 2015-08-06
Fix imports order to match the new Checstyle rule

ppalaga:


ppalaga:

PR #34, opened 2015-08-10
HAWKULAR-530 pass around paths not IDs

jmazzitelli:


jmazzitelli:
Do not merge/release until the UI is changed so it passes resource paths, and not IDs.

PR #36, opened 2015-08-11
HWKAGENT-5 resource config in inventory

jmazzitelli:
HawkularStorageAdapter.java line 276 and following needs to be changed to use the new inventory API. Waiting on inventory REST API to be implemented. See HWKINVENT-113

PR #38, opened 2015-08-21
escaping character issues

jkremser:
**This is a sub set of the #36**

This change is needed to work properly with inventory 0.3.0, otherwise the discovery is stopped after first failure that happens because of the way how agent escapes the slashes... it was changed in the 0.3.0, (no more '\/').


jmazzitelli:
I just manually merged this

PR #43, opened 2015-09-07
Use cmdgw rather than feedcomm in Java package names

ppalaga:


ppalaga:
Depends on https://github.com/hawkular/hawkular-command-gateway/pull/1

PR #45, opened 2015-09-09
Enable container-based infrastructure and cache ~/.m2/repository for faster Trav

ppalaga:


ppalaga:
@jmazzitelli I after all came to the conclusion that this is a better home for the cmdgw itests. First, adding them to hawkular looked like too much overhead for Travis (log length over limit, diskspace problems, making the build slower) and second this is a little bit closer to the subject under test.

An apparent disadvantage of this approach is that we will have to maintain the distro for the tests both here and in hawkular.

Could you please review and eventually merge?

PR #48, opened 2015-09-11
Add AddDatasourceCommand

ppalaga:


hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/inventory/InventoryIdUtil.java
```diff
@@ -25,6 +25,16 @@
...
+
+        public String getFeedId() {
+            return feedId;
+        }
+        public String getManagedServerName() {
+            return managedServerName;
+        }
+        public String getIdPart() {
+            return idPart;
+        }
```
jmazzitelli, line 37: if you want to make public getters, then make the data fields private so it forces people to go through the getters.

ppalaga, line 37: Yeah, wanted to do that in a separate commit, but... forgot. Will do.

PR #54, opened 2015-09-18
Add driverClass, driverMajorVersion and driverMinorVersion to AddJdbcDriverReque

ppalaga:


ppalaga:

hawkular-dmr-client/src/main/java/org/hawkular/dmrclient/DatasourceJBossASClient.java
```diff
@@ -29,6 +30,7 @@
  * @author John Mazzitelli
  */
 public class DatasourceJBossASClient extends JBossASClient {
+    private static final Logger log = Logger.getLogger(DatasourceJBossASClient.class);
```
jmazzitelli: cant' you juse use the protected "log" member from JBossASClient - that's what its intended purpose was.

ppalaga: Call me stupid, I have not noticed that one. Let me fix it. Thanks

hawkular-dmr-client/src/main/java/org/hawkular/dmrclient/DatasourceJBossASClient.java
```diff
@@ -152,33 +152,46 @@ public boolean isXADatasource(String datasourceName) throws Exception {
     /**
      * @param driverName the name of the JDBC driver to create, e.g. {@code "h2"}
      * @param moduleName the name of the WIldFly module that contains the driver jar file
+     * @return
```
jmazzitelli: empty return javadoc will fail javadoc verifier (if we ever run that). just put something in there after "return" to minimally describe it

PR #56, opened 2015-09-21
HAWKULAR-609 - Implemented the 'Export JDR' operation.

jpkrohling:


hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/feedcomm/ExportJdrCommand.java
```diff
@@ -0,0 +1,152 @@
...
+        CanonicalPath canonicalPath = CanonicalPath.fromString(request.getResourcePath());
+        String resourceId = canonicalPath.ids().getResourcePath().getSegment().getElementId();
+        InventoryIdUtil.ResourceIdParts idParts = InventoryIdUtil.parseResourceId(resourceId);
+        ManagedServer managedServer = config.managedServersMap.get(new Name(idParts.getManagedServerName()));
+        if (managedServer == null) {
+            throw new IllegalArgumentException(String.format("Cannot execute operation: unknown managed server [%s]",
+                    idParts.getManagedServerName()));
+        }
+
+        if (managedServer instanceof LocalDMRManagedServer || managedServer instanceof RemoteDMRManagedServer) {
```
pilhuhn, line 73: Isn't that always true?

ppalaga:
Finished the first review round

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/feedcomm/ExportJdrCommand.java
jmazzitelli, line 73: It is today, but not in the future. This is just code for the future - later on when we introduce JMS-based managed servers, we'll want to do something different for those.

PR #67, opened 2015-09-29
HWKAGENT-10 add new dependencies that are being pulled in via inventory

jmazzitelli:


hawkular-wildfly-monitor/src/main/resources/module/main/module.xml
```diff
@@ -34,7 +34,25 @@
     <resource-root path="jackson-annotations.jar"/>
     <resource-root path="jackson-core.jar"/>
     <resource-root path="jackson-databind.jar"/>
+
+    <!-- These are pulled in due to hawkular-inventory-json-helper.jar.
+         It is probable that we really don't need all of these,
+         but without knowing which are really used by hawkular-inventory-json-helper
+         it is best we include all of them. -->
```
pilhuhn, line 41: I am not happy about that. 
We should try to keep the agent as small as possible.

jmazzitelli, line 41: We need to ask the inventory folks what, if any, of these are really needed. They should add exclusions to their mvn artifact dependencies  (not sure if the agent could pick up those exclusions via transitive dependency rules or if the agent pom itself would need to add exclusions as well).

PR #72, opened 2015-10-05
Combination of mine + mazz's work towards a feature-pack.zip

bobmcwhirter:
This may or may not actually work, and given that we have multiple module.xml's to keep in sync, it's sub-optimal.  But it doesn't break the build.  At least on my machine.


bobmcwhirter:
Adjusted the primary module.xml to use the jackson module's that exist in wildfly, instead of pulling in the .jars as local <resource>s

PR #73, opened 2015-10-06
Use bulk add operations in inventory

ppalaga:

PR #74, opened 2015-10-07
using inventory batch API with queue buffering and diagnostics

jmazzitelli:
this incorporates but also replaces PR #73 - see commit messages for details


ppalaga:
... and I was hoping to silently get rid of those useless multiline diagnostic messages in the log ;)

jmazzitelli:
Hardly useless! Very helpful in diagnosing performance! Take a closer look at the information in them. BTW: those diagnostic metrics can actually be stored in hawkular-metrics itself rather than in the server log (so the agent can store its own metrics in h-metrics) by changing "LOG" to "STORAGE" in the <diagnostics> configuration in standalone.xml.

jmazzitelli:
Oh, and you can also change the interval at which those diagnostic logs are emitted - again, configurable in <diagnostics> element in standalone.xml. Right now I believe it is set at 1 minute, but you can set it to anything (5 minutes, 60 minutes, whatever)

ppalaga:
Configurable? - Great! I'll set it to max in my env not to see it ;)

jmazzitelli:
You can set it to 1,000,000 minutes if you never want to see it :)

PR #75, opened 2015-10-07
Implement remove operations on Datasources and JDBC drivers

ppalaga:


ppalaga:
@jmazzitelli the remove ops are not fully polished yet - they just remove the DMR node, leaving filesystem objects as is. My plan is to fix that in the next PR. I am sending it in this form so that @ammendonca has the API to write UI against.

@jmazzitelli, could you please review?

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/cmd/AbstractRemoveModelNodeCommand.java
```diff
@@ -0,0 +1,144 @@
...
+     * @param managedServerName
+     * @param managedServer
+     */
+    protected void validate(REQ request, String managedServerName, ManagedServer managedServer) {
+        if (managedServer == null) {
+            throw new IllegalArgumentException(
+                    String.format("Cannot add [%s]: unknown managed server [%s]", entityType, managedServerName));
+        }
+
+        if (!(managedServer instanceof LocalDMRManagedServer)) {
```
jmazzitelli: I see this only works when the managed server is the local WF server - this should be loudly advertised in this abstract class' javadoc so users of this class know this.

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/cmd/AbstractRemoveModelNodeCommand.java
```diff
@@ -0,0 +1,144 @@
...
+        CanonicalPath canonicalPath = CanonicalPath.fromString(rawResourcePath);
+
+        String resourceId = canonicalPath.ids().getResourcePath().getSegment().getElementId();
+        ResourceIdParts idParts = InventoryIdUtil.parseResourceId(resourceId);
+        String modelNodePath = idParts.getIdPart();
+
+        RESP response = createResponse();
+        MessageUtils.prepareResourcePathResponse(request, response);
+
+        DatasourceJBossASClient dsc = null;
```
jmazzitelli: I see this  AbstractRemoveModelNodeCommand class is abstract and thus I assume meant for reuse - however, it is using the very datasource-specific class of DatasourceJBossASClient. Why is that?

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/cmd/AbstractRemoveModelNodeCommand.java
```diff
@@ -0,0 +1,144 @@
...
+import org.hawkular.cmdgw.api.ResourcePathResponse;
+import org.hawkular.cmdgw.api.ResponseStatus;
+import org.hawkular.dmrclient.Address;
+import org.hawkular.dmrclient.DatasourceJBossASClient;
+import org.hawkular.inventory.api.model.CanonicalPath;
+
+/**
+ * @author <a href="https://github.com/ppalaga">Peter Palaga</a>
+ */
+public abstract class AbstractRemoveModelNodeCommand<REQ extends ResourcePathRequest, RESP extends ResourcePathResponse>
```
jmazzitelli, line 42: since this is an abstract class meant for reuse, fill in the javadocs with the details so developers know how to use this and what its for.

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/cmd/AbstractRemoveModelNodeCommand.java
```diff
@@ -0,0 +1,144 @@
...
+        DatasourceJBossASClient dsc = null;
+        try {
+            String managedServerName = idParts.getManagedServerName();
+            ManagedServer managedServer = config.managedServersMap.get(new Name(managedServerName));
+            validate(request, managedServerName, managedServer);
+
+            validate(modelNodePath, request);
+            DMRInventoryManager inventoryManager = context.getDiscoveryService().getDmrServerInventories()
+                    .get(managedServer);
+            dsc = new DatasourceJBossASClient(inventoryManager.getModelControllerClientFactory().createClient());
```
jmazzitelli: If the reason the use of the datasource client is to use the .remove API, I would recommand using CoreJBossASClient to avoid the impression that the datasource functionality is needed. the "core" client's purpose was to provide functionality to the core server capabilities - this seems more appropriate to use than the datasource one. if anything this would avoid confusion to someone reading the code wondering why the datasource client is being used.

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/cmd/AbstractRemoveModelNodeCommand.java
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        return new BasicMessageWithExtraData<>(response, null);
+
+    }
+
+    /**
+     * @param modelNodePath
+     * @param request
+     */
```
jmazzitelli, line 114: fill in javadoc to explain what this is validating  - same with the empty javadocs in the method below

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/cmd/AbstractRemoveModelNodeCommand.java
ppalaga: Actually a good catch. I just realized that Add/Remove Datasource is a DMR-only operation, so this needs to be different for DS and drivers. Let me fix it.

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/cmd/AbstractRemoveModelNodeCommand.java
ppalaga: A good catch again :) `JBossASClient` will be enough.

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/cmd/AbstractRemoveModelNodeCommand.java
ppalaga: Yes, fixed in 18ece02

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/cmd/AbstractRemoveModelNodeCommand.java
ppalaga, line 114: Fixed in 18ece02

ppalaga:
18ece02 adds JavaDoc, replaces `DatasourceJBossASClient` with `JBossASClient` and fixes the server type validation constraints.

PR #76, opened 2015-10-09
Update Datasource command

ppalaga:


ppalaga:
@jmazzitelli this is not fully polished yet, but I prefer to show you some basic things before proceeding so that you have a chance to stop me early :)

(1) 43830c0 changes the signature of `Command.execute()` to use `BasicMessageWithExtraData<REQ>` rather than `REQ` and `BinaryData`. This makes it easier put the parts common for all commands into `AbstractResourcePathCommand.execute()`

(2) 2b7b065 Fluent API for WF management operations. I find this kind of design easier to use and maintain than the `*Client` classes we have in dmrclient now. I have not looked over all available clients, but I think I am not far away from being able to replace all of them with the new fluent code, which I'd like to do at least for all Datasource and Driver commands.

Please comment.

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
```diff
@@ -0,0 +1,446 @@
...
+        private AddOperationBuilder(BatchOperationBuilder<BatchOperationBuilder<?>> batch) {
+            super(batch, ADD);
+        }
+
+        /**
+         * If {@code value == null}, no entry is added. Otherwise, the attribute will be added.
+         *
+         * @param name
+         * @param value
+         * @return
```
jmazzitelli: this will fail javadoc validator if its run - fill in @param and @return or remove them.

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/cmd/FeedCommProcessor.java
```diff
@@ -243,13 +246,13 @@ public void onClose(int reasonCode, String reason) {
...
         if (!(disconnectCode == reasonCode && disconnectReason.equals(reason))) {
             switch (reasonCode) {
-                case 1008: { // VIOLATED POLICY - don't try again since it probably will fail again (bad credentials?)
-                    break;
-                }
-                default: {
-                    startReconnectJobThread();
-                    break;
-                }
+            case 1008: { // VIOLATED POLICY - don't try again since it probably will fail again (bad credentials?)
```
jmazzitelli: hawkular coding standards indent case stmts inside switch - I just double checked and alerts and metric code as well a bus all indent case stmts inside switch (i.e. case and switch are not on the same indent level).

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/cmd/FeedCommProcessor.java
```diff
@@ -278,20 +281,19 @@ public void onMessage(BufferedSource payload, WebSocket.PayloadType payloadType)
...
-                    case BINARY: {
-                        InputStream input = payload.inputStream();
-                        msgWithData = new ApiDeserializer().deserialize(input);
-                        break;
-                    }
-                    default: {
-                        throw new IllegalArgumentException("Unknown payload type, please report this bug: "
-                                + payloadType);
-                    }
+                case TEXT: {
```
jmazzitelli: case stmts should be indented

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/cmd/UpdateDatasourceCommand.java
```diff
@@ -0,0 +1,178 @@
...
+            batch.add() //
+                    .address().segments(parentAddress).segment(propType, newPropName).parentBuilder() //
+                    .valueAttribute(newProp.getValue()).parentBuilder();
+        }
+    }
+
+    /**
+     * @param adr
+     * @param datasourceName
+     */
```
jmazzitelli: delete this empty javadoc

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
ppalaga: Holly yes, I know you see everything, which is one of the reasons I have put [Do not merge yet] on this PR. I should turn that stupid validator on (in my workspace at least) not to bother you with such stupidities. But thinking of it once again, your care for these nits might also be interpreted as a sign that you like the underlying code(?)

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/DmrConstants.java
```diff
@@ -0,0 +1,483 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.hawkular.dmr.api;
+
+/**
+ * @author <a href="https://github.com/ppalaga">Peter Palaga</a>
+ */
+public interface DmrConstants  {
```
jmazzitelli: where did these DMR constants come from? Do we need all of these?

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/cmd/FeedCommProcessor.java
ppalaga: That's a new piece info for me. thanks :)

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
```diff
@@ -0,0 +1,446 @@
...
+        }
+
+        public T valueAttribute(String value) {
+            return attribute(VALUE, value);
+        }
+
+    }
+
+    public static class AddressBuilder<T extends AddressBuilder<?, P>, P extends AbstractSingleOperationBuilder<?, ?>> {
+        private final ModelNode baseNode = new ModelNode();
```
jmazzitelli, line 180: curious why you didn't use the Address DMR client class - seems like that could have been useful here

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/DmrConstants.java
ppalaga: Brutally copied from `org.jboss.as.controller.descriptions.ModelDescriptionConstants`. We could perhaps consume them also from there, wdyt?

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
```diff
@@ -0,0 +1,446 @@
...
+import java.util.StringTokenizer;
+
+import org.jboss.as.controller.client.ModelControllerClient;
+import org.jboss.dmr.ModelNode;
+import org.jboss.dmr.Property;
+
+/**
+ * @author <a href="https://github.com/ppalaga">Peter Palaga</a>
+ */
+public class OperationBuilder implements DmrConstants, SubsystemDatasourceConstants, SubsystemLoggingConstants {
```
jmazzitelli: I think you really need to javadoc all these static methods here - what are they used for, what do they give you, even how are they used. Right now, since there is so much generic typing going on in here, its hard to understand the overall usage of these methods. Commenting this code would be very helpful for people who could use this.

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
ppalaga, line 180: Yes, I could use it like this https://github.com/hawkular/hawkular-agent/pull/76/files#diff-16951fe3c9c943afaf15fa9b84b94f9bR94 but not like this https://github.com/hawkular/hawkular-agent/pull/76/files#diff-bd3cab1f7c8cb010c57a853029b9568dR137 or even like this https://github.com/hawkular/hawkular-agent/pull/76/files#diff-3e0d35c6e0bc4e562516e3474d83939eR434 - i.e. as a part of a larger fluent expression.

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
ppalaga: Well, i did not want to invest too much time into JavaDoc before I get some kind of approval for this kind of design from you. I am ready to document it thoroughly after that. To get a good impression, what these builders are good for, just search for occurences of `OperationBuilder`.

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/DmrConstants.java
jmazzitelli: I think we can consume them from there rather than directly copying them. IIRC, we'll have that class on the classpath (I believe we have the controller API since it should come with the controller dependency declared in the dmr-client pom - if that is true, we should just use them directly and not copy them).

ppalaga: OK, let's consume their's `ModelDescriptionConstants`. It is not an interface as mine `DmrConstants` to be able to "import via implements" but nevermind. 

Thinking again about possible future changes in `ModelDescriptionConstants`, they'll most probably just be adding new constants and never remove one, which is something we can live with in the long term. I mean there probably comes a day when we'll have to support multiple EAP/WF versions and we'll be able to import the `ModelDescriptionConstants` just from the newest AS. WF core team should not cause any problems to us if they behave responsibly.

ppalaga:
This got mergeable from my PoV. @jmazzitelli could you please review?

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
```diff
@@ -0,0 +1,751 @@
...
+ * &#47;&#47; remove a datasource with another way of providing an address
+ * OperationBuilder.remove()
+ *          .address().segments("/subsystem-datasources/data-source=myDatasource").parentBuilder()
+ *          .execute(mcc)
+ *          .assertSuccess();
+ * </pre>
+ *
+ * @author <a href="https://github.com/ppalaga">Peter Palaga</a>
+ */
+public class OperationBuilder implements SubsystemDatasourceConstants, SubsystemLoggingConstants {
```
jmazzitelli, line 57: it seems odd this is a general-use builder, but it has only datasource constants that it implements. I don't think this is only for datasources, but why is it implementing SubsystemDatasourceConstants? Is this only for datasources?

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
jmazzitelli: clean up empty tags

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
```diff
@@ -0,0 +1,751 @@
...
+            return (T) this;
+        }
+
+        /**
+         * If {@code value == null}, no entry is added. Otherwise, the attribute will be added.
+         *
+         * @param name
+         * @param value
+         * @return
+         */
```
jmazzitelli, line 164: clean up empty tags

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
```diff
@@ -0,0 +1,751 @@
...
+                }
+            }
+            return (T) this;
+        }
+
+        public T subsystem(String subsystem) {
+            return segment(ModelDescriptionConstants.SUBSYSTEM, subsystem);
+        }
+
+        public T subsystemDatasources() {
```
jmazzitelli, line 238: again, seems odd that only datasource stuff is here. but I guess this is just for convenience since you use ths in the datasource stuff.

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
```diff
@@ -0,0 +1,751 @@
...
+        protected final ModelNode responseNode;
+
+        private OperationResult(ModelNode requestNode, ModelNode responseNode) {
+            super();
+            this.requestNode = requestNode;
+            this.responseNode = responseNode;
+        }
+
+        /**
+         * Asserts that the {@link ModelDescriptionConstants#OUTCOME} fielf of {@link #responseNode} is equal to
```
jmazzitelli: typo - "field"

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
```diff
@@ -0,0 +1,751 @@
...
+            super(bb, ModelDescriptionConstants.WRITE_ATTRIBUTE_OPERATION);
+        }
+
+        /**
+         * If {@code value == null}, the attribute will be removed. Otherwise, the attribute will be written.
+         *
+         * @param name
+         * @param value
+         * @return
+         */
```
jmazzitelli, line 686: fix empty tags

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/cmd/UpdateDatasourceCommand.java
```diff
@@ -0,0 +1,236 @@
...
+        case XA_DATASOURCE:
+            return true;
+        case DATASOURCE:
+            return false;
+        default:
+            String msg = String.format(
+                    "DMR Address [%s] was supposed to be either a [%s] or [%s] address, which it is not",
+                    adr.toString(), XA_DATASOURCE, DATASOURCE);
+            throw new IllegalStateException(msg);
+        }
```
jmazzitelli, line 157: fix case indentation

jmazzitelli:
I had a minor question or two and we also need to fix those nitpicks of mine (mainly around empty javadoc tags and indentation). but once those are addressed I think this can get merged

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
ppalaga, line 57: > Is this only for datasources?

No, it is general purpose. `implements` is just "misused" to import all constants from the interface. 

If your objection is that a general-purpose builder should not rely on/use subsystem specific constants, then... well... that's a good question :) I think it is OK to have names of subsystems here, because the set of subsystems is fairly stable. And having it here, we can offer handy shortcuts like `AddressBuilder.subsystemDatasources()` https://github.com/hawkular/hawkular-agent/pull/76/files#diff-650a4b8ae9cc61f2557a8bed543e3087R238

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
ppalaga, line 238: Right, this is for the convenience.

ppalaga:
21c9b18 should fix all requests

PR #77, opened 2015-10-11
this removes the "double send" of metrics.

jmazzitelli:
once hwkmetrics 83 is fixed, we need to incorporate this in the release


jmazzitelli:
HWKAGENT-83 is resolved, we can remove the "two messages per metric" from the agent now

PR #80, opened 2015-10-13
HAWKULAR-650 - Provide (remote) agent install(er) for WF-agent

lzoubek:
This PR adds ability to install hawkular-widfly-agent to a widlfly server.

Hawkular Widlfly agent  module packaging has been changed, so it includes subsystem-snippet.xml within module zip. extension-installer can find snippet and add such subsystem configuration into standalone.xml

Agent installer does not bundle agent module within itself. It's capable of downloading agent module from hawkular server (module location needs to be fixed, and hawkular server needs to provide such file) or can install module usin `--module` command-line option.

extension-installer even supports installation to domain - but it's not used yet.

### DIFF
--- a/hawkular-wildfly-monitor/src/main/assembly/module-assembly.xml
+++ b/hawkular-wildfly-monitor/src/main/assembly/module-assembly.xml
@@ -39,6 +39,14 @@
       <directoryMode>0755</directoryMode>
     </fileSet>
   </fileSets>
+  <files>
+    <file>
+      <source>${project.basedir}/src/main/assembly/subsystem.xml</source>
+      <outputDirectory>/${moduleDir}/main</outputDirectory>
+      <destName>subsystem-snippet.xml</destName>
+    </file>
+  </files>
+
 
   <dependencySets>
     <dependencySet>
--- a/pom.xml
+++ b/pom.xml
@@ -39,6 +39,8 @@
     <module>hawkular-wildfly-monitor</module>
     <module>hawkular-wildfly-monitor-feature-pack</module>
     <module>hawkular-agent-itest-parent</module>
+    <module>wildfly-module-installer</module>
+    <module>hawkular-wildfly-monitor-installer</module>
   </modules>
 
   <scm>
--- a/hawkular-wildfly-monitor-installer/README.adoc
+++ b/hawkular-wildfly-monitor-installer/README.adoc
@@ -1,0 +1,35 @@
+= Hawkular Wildfly Monitor Installer
+:source-language: java
+
+== About
+
+Hawkular Wildfly Monitor Installer is a standalone java application to install Hawkular Agent into stock Wildfly Server.
+
+== Running using Maven
+
+Installer is configured via maven exec plugin so it installs Hawkular Agent from this repository into local Wildfly located in `${wildfly.home}`
+
+    mvn exec:java -Dwildfly.home=/opt/wildfly-9.0.0.Final
+
+== Running as standalone java application
+
+=== Building standalone jar
+
+    mvn package assembly:single
+
+=== Running
+
+Command line parameter `module` is required unless installer retrieves Hawkular Monitor module from `hawkular-server-url`
+
+   java -jar target/hawkular-wildfly-monitor-installer-0.10.1.Final-SNAPSHOT-standalone.jar \
+     --module=/path/to/module.zip \
+     --wildfly-home=/opt/wildfly-9.0.0.Final \
+     --hawkular-server-url=http://localhost:8080
+
+Hawkular Agent is then installed to Wildfly Server `/opt/wildfly-9.0.0.Final` as a module and is able to report to Hawkular Server located at `http://localhost:8080`.
+
+=== Setting up SSL
+
+If needed, installer can setup secured connection between Hawkular Wildfly Monitor and Hawkular server. By specifying `https` scheme  (ie. `--server-url=https://localhost:8443`)
+following command-line options become required `--keystore-path` and `--key-alias`. Installer also needs `--keystore-password` and `--key-password` options
+but they can be omitted in case you run it in console (in such case, installer will prompt for passwords). For more details see http://www.hawkular.org/docs/user/secure-comm.html[Secure Communications].
--- a/hawkular-wildfly-monitor-installer/pom.xml
+++ b/hawkular-wildfly-monitor-installer/pom.xml
@@ -1,0 +1,87 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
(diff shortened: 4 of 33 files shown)


pom.xml
```diff
@@ -39,6 +39,8 @@
     <module>hawkular-wildfly-monitor</module>
     <module>hawkular-wildfly-monitor-feature-pack</module>
     <module>hawkular-agent-itest-parent</module>
+    <module>extension-installer</module>
```
jmazzitelli: rename this module something more descriptive like "wildfly-module-installer" or "wildlfy-subsystem-extension-installer" ... "extension-installer" has no indication that its for wildfly

pom.xml
```diff
@@ -39,6 +39,8 @@
...
+    <module>hawkular-agent-installer</module>
```
jmazzitelli: rename this to match the name we use elsewhere "hawkular-wildfly-monitor" - so call this module something like "hawkular-wildfly-monitor-installer"

jmazzitelli:
due to recent additions to the agent (specifically about supporting SSL/https communications), we need to be able to optionally configure SSL in standalone.xml. Can this new code be used to add things like &lt;security-realm> and &lt;https-listener>? See https://github.com/hawkular/hawkular.github.io/blob/pages/src/main/jbake/content/docs/user/secure-comm.adoc for what we would need to be added into standalone.xml.

hawkular-wildfly-monitor/src/main/assembly/module-assembly.xml
```diff
@@ -39,6 +39,14 @@
...
+    <file>
+      <source>${project.basedir}/src/main/assembly/subsystem.xml</source>
+      <outputDirectory>/${moduleDir}</outputDirectory>
+      <destName>subsystem-snippet.xml</destName>
+    </file>
+  </files>
+
 
   <dependencySets>
     <dependencySet>
```
jmazzitelli, line 52: will this new snippet file cause problems if someone wants to manually install the module by just unzipping in the WildFly modules/ directory? We don't want to FORCE someone to have to use this installer to install our special module .zip file.

jmazzitelli, line 52: Does this affect in a any way the new feature pack build? We want to make sure this doesn't break or otherwise affect the feature pack and the ability of swarm to consume our feature pack. (see the feature pack module in our repo).

pilhuhn, line 52: This is a bit tricky: either we supply that file in the installer, but then we need to update the installer when it changes. Or the installer needs to unpack the module.zip!org/hawkular/agent/monitor/main/hawkular-monitor.jar to find it. Or we require downloading from a running (web) server which then needs to serve that. 
For automation purposes, I think the possibility for offline install is valid.

pilhuhn:
@jmazzitelli I'd like to see that included as is (with the smaller changes you suggested) and then being worked on as we continue fleshing out more details. Also when it is in the main repo, we have possibly more eyes to look at.
I can confirm that Libors code nicely does what it is supposed to do with wf10 servers.

pom.xml
lzoubek: ok

lzoubek:
Yes,

extension-installer is capable of doing it, right now it configures  
<storage-adapter similar way (although <storage-adapter comes from  
subsystem-snippet.xml)

here  
https://github.com/hawkular/hawkular-agent/pull/80/files#diff-bd9480e4c46e1902f1316bcf9c415a6fR86

in above case the content is taken as generated String, but it can also be  
read from File or (I'd need to code it) URL (JAR resource)

On Tue, 13 Oct 2015 14:49:04 +0200, John Mazzitelli  
notifications@github.com wrote:

> due to recent additions to the agent (specifically about supporting  
> SSL/https communications), we need to be able to optionally configure  
> SSL in standalone.xml. Can this new code be used to add things like  
> <security-realm> and <https-listener>? See  
> https://github.com/hawkular/hawkular.github.io/blob/pages/src/main/jbake/content/docs/user/secure-comm.adoc  
> for what we would need to be added into standalone.xml.
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/hawkular/hawkular-agent/pull/80#issuecomment-147705197

## 

Libor Zoubek

hawkular-wildfly-monitor/src/main/assembly/module-assembly.xml
lzoubek, line 52: I think it won't cause issues as it is not referenced in module.xml, but it's maybe better to place it deeper into ZIP right next to module.xml

hawkular-agent-installer/src/main/java/org/hawkular/agentinstaller/AgentInstaller.java
```diff
@@ -0,0 +1,169 @@
...
+     * @param hawkularServerUrl
+     * @return
+     */
+    private static XmlEdit replaceServerUrl(String hawkularServerUrl) {
+        String select = "/server/profile/"
+                + "*[namespace-uri()='urn:org.hawkular.agent.monitor:monitor:1.0']/";
+
+        StringBuilder xml = new StringBuilder("<storage-adapter")
+            .append(" type=\"HAWKULAR\"")
+            .append(" url=\""+hawkularServerUrl+"\"")
```
jmazzitelli: we do NOT want to use "url" attribute (I'm gonna get rid of this soon since everyone keep using it :)) - we want to use the standard binding mechanism that WF provides; putting in a url with custom ports violates standard practice in WF subsystems. Instead, we want to add a &lt;outbound-socket-binding that defines the host/port to the hawkular server:

&lt;outbound-socket-binding name="hawkular">
  <&lt;remote-destination host="your-hawkular-server-hostname"  port="the-port" />
&lt;/outbound-socket-binding>

This is why i ask if we can use this tool to add extra stuff in standalone.xml. We need to add this outbound socket binding, and optionally the security-realm and https-listener.

lzoubek:
@jmazzitelli please do not merge yet, I'd like to squash all the stuff into 1 commit, once you accept this PR

jmazzitelli:
would you consider this done? does it add optional security-realm and https-listener elements to standalone.xml (to support https/ssl)?

lzoubek:
@jmazzitelli I added SSL support to installer, it will setup security-realm and storage-adapter in case it detects https:// in `--hawkular-server-url`

Is `https-listener` also required (it is not clear to me from documentation)

jmazzitelli:
> Is https-listener also required (it is not clear to me from documentation)

No, that is only for kettle (the hawkular server)

PR #85, opened 2015-10-16
Cleanup itests 

ppalaga:


pom.xml
```diff
@@ -78,6 +78,12 @@
       </dependency>
 
       <dependency>
+        <groupId>commons-io</groupId>
+        <artifactId>commons-io</artifactId>
+        <version>2.4</version>
+      </dependency>
```
jmazzitelli, line 84: don't forget that any new dependencies have to be added to modules.xml in both the agent assembly and in the feature pack module.xml

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
```diff
@@ -654,19 +654,19 @@ private StringListOperationResult(ModelNode requestNode, ModelNode resultNode) {
         }
 
         public Set<String> getHashSet() {
-            return Collections
-                    .unmodifiableSet(getNodeList().stream().map(n -> n.asString()).collect(Collectors.toSet()));
+            Set<String> result = getNodeList().stream().map(n -> n.asString()).collect(Collectors.toSet());
```
jmazzitelli, line 657: this still requires Java 8 correct?

ppalaga, line 657: Yes, let's speak about the Java7 support on monday.

PR #87, opened 2015-10-17
HAWKULAR-604 The message field of ExecuteOperationResponse should not…

josejulio:
… be "undefined"

Refactored the ExecuteOperationCommand to extend from AbstractResourcePath and to use OperationBuilder.
Updated ExecuteOperationCommandITest to reflect the refactoring.
Made a GenericOperationBuilder class to create generic operations.

### DIFF
--- a/hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
+++ b/hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
@@ -715,6 +715,31 @@ public class OperationBuilder implements SubsystemDatasourceConstants, Subsystem
 
     }
 
+    public static class ByNameOperationBuilder <T extends ByNameOperationBuilder<?>>
+            extends AbstractSingleOperationBuilder<T, OperationResult<?>> {
+
+        public ByNameOperationBuilder(CompositeOperationBuilder<CompositeOperationBuilder<?>> bb,
+                String operationName) {
+            super(bb, operationName);
+        }
+
+        /**
+         * If {@code value == null}, no entry is added. Otherwise, the attribute will be added.
+         *
+         * @param name the name of the attribute
+         * @param value the value of the attribute
+         * @return this builder
+         */
+        @SuppressWarnings("unchecked")
+        public T attribute(String name, String value) {
+            if (value != null) {
+                baseNode.get(name).set(value);
+            }
+            return (T) this;
+        }
+
+    }
+
     private static final MsgLogger log = DmrApiLoggers.getLogger(OperationBuilder.class);
 
     public static AddOperationBuilder<AddOperationBuilder<?>> add() {
@@ -765,4 +790,8 @@ public class OperationBuilder implements SubsystemDatasourceConstants, Subsystem
         return new WriteAttributeOperationBuilder<>(null);
     }
 
+    @SuppressWarnings("unchecked")
+    public static ByNameOperationBuilder<ByNameOperationBuilder<?>> byName(String operationName) {
+        return new ByNameOperationBuilder<>(null, operationName);
+    }
 }
--- a/hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/cmd/ExecuteOperationCommand.java
+++ b/hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/cmd/ExecuteOperationCommand.java
@@ -19,68 +19,66 @@ package org.hawkular.agent.monitor.cmd;
 import java.util.Collection;
 import java.util.Map;
 
-import org.hawkular.agent.monitor.extension.MonitorServiceConfiguration;
 import org.hawkular.agent.monitor.inventory.ID;
-import org.hawkular.agent.monitor.inventory.InventoryIdUtil;
-import org.hawkular.agent.monitor.inventory.InventoryIdUtil.ResourceIdParts;
 import org.hawkular.agent.monitor.inventory.ManagedServer;
-import org.hawkular.agent.monitor.inventory.Name;
 import org.hawkular.agent.monitor.inventory.ResourceManager;
 import org.hawkular.agent.monitor.inventory.dmr.DMRInventoryManager;
 import org.hawkular.agent.monitor.inventory.dmr.DMROperation;
 import org.hawkular.agent.monitor.inventory.dmr.DMRResource;
-import org.hawkular.agent.monitor.inventory.dmr.LocalDMRManagedServer;
-import org.hawkular.agent.monitor.inventory.dmr.RemoteDMRManagedServer;
 import org.hawkular.agent.monitor.log.AgentLoggers;
 import org.hawkular.agent.monitor.log.MsgLogger;
 import org.hawkular.bus.common.BasicMessageWithExtraData;
 import org.hawkular.cmdgw.api.ExecuteOperationRequest;
 import org.hawkular.cmdgw.api.ExecuteOperationResponse;
-import org.hawkular.cmdgw.api.MessageUtils;
-import org.hawkular.cmdgw.api.ResponseStatus;
+import org.hawkular.dmr.api.OperationBuilder;
 import org.hawkular.dmrclient.Address;
-import org.hawkular.dmrclient.CoreJBossASClient;
-import org.hawkular.dmrclient.JBossASClient;
 import org.hawkular.inventory.api.model.CanonicalPath;
 import org.jboss.as.controller.client.ModelControllerClient;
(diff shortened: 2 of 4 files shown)


pilhuhn:
@josejulio Thank you for your contribution - the Travis failure seems transient - I've kicked of a 2nd build that passes. I am sure @ppalaga will have a look at your patch on Monday.

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
```diff
@@ -765,4 +790,9 @@ public T valueAttribute(String value) {
         return new WriteAttributeOperationBuilder<>(null);
     }
 
+    @SuppressWarnings("unchecked")
+    public static GenericOperationBuilder<GenericOperationBuilder<?>> generic(String operationName) {
```
jmazzitelli: this method name is too "generic" :) As a user of this API, I have no idea what this means - and with no javadoc, it isn't clear what it is. I assume by the name of the parameter, this defines the DMR operation name that is to be executed. Or does it create a full DMR ModelNode that is to be used to execute an operation? Couldn't this method be called "operation" or "executeOperation" or "operationName" to be explicit what its for?

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/cmd/ExecuteOperationCommand.java
```diff
@@ -19,68 +19,61 @@
...
+    public ExecuteOperationCommand() {
+        super("Execute Operation", "DMR Node");
+    }
+
+    /** @see org.hawkular.agent.monitor.cmd.AbstractResourcePathCommand#createResponse() */
     @Override
-    public BasicMessageWithExtraData<ExecuteOperationResponse> execute(
-            BasicMessageWithExtraData<ExecuteOperationRequest> envelope, CommandContext context) throws Exception {
-        ExecuteOperationRequest request = envelope.getBasicMessage();
-        log.infof("Received request to execute operation [%s] on resource [%s]", request.getOperationName(),
```
jmazzitelli, line 59: now that we refactored this to use the generic operation execution builder, when is this message logged? I still want to see in the log message that a request was received to execute operation O on resource R. For support and debugging purposes we need to be able to know that the agent received this request. I looked but didn't see any log messages in the new OperationBuilder  - is it logged somewhere else?

josejulio:
@pilhuhn Thank you, I though i did something wrong.

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
josejulio: Oh! The "generic" irony :).

It defines the operation name, useful when you don't know exactly the operation that is going to be executed until runtime.

I chose "generic" to follow other API calls:

| Builder | method |
| --- | --- |
| AddOperationBuilder | add() |
| MapPutOperationBuilder | mapPut() |
| ReadResourceOperationBuilder | readResource() |
| GenericOperationBuilder | generic() |

However, I could change the name and/or add the javadoc to make it more helpful.

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/cmd/ExecuteOperationCommand.java
josejulio, line 59: This (an example on running the tests on my computer):

```
Received request to perform [Execute Operation] on a [DMR Node] given by inventory path [/t;28026b36-8fe4-4332-84c8-524e173a68bf/e;test/f;earthquake/r;earthquake~Local~~/r;earthquake~Local~%2Fdeployment=hawkular-helloworld-war.war]
```

is written here: https://github.com/hawkular/hawkular-agent/blob/master/hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/cmd/AbstractResourcePathCommand.java#L194-L197

Is that enough? or should i return the log to the ExecuteOperationCommand?

``` java
log.infof("Received request to execute operation [%s] on resource [%s]", request.getOperationName(),
```

Or maybe we could make these attributes not final to change the [Execute Operation] and [DMR Node] on runtime.
https://github.com/hawkular/hawkular-agent/blob/master/hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/cmd/AbstractResourcePathCommand.java#L51-L62

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
ppalaga: @jmazzitelli that `generic()` name was my spontaneous suggestion and I do not find it ideal myself. Let's find a better name.

As @josejulio points out, the sibling methods in this class are called 1:1 after the operation whose builder they return. This one returns a generic operation builder by name. 

> Couldn't this method be called "operation" or "executeOperation" or "operationName" to be explicit what its for?

From your proposals, I find `operationName` the best. 

My proposals are `OperationBuilder.byName("reload")...` or `OperationBuilder.byOperationName("reload")...` 
WDYT?

josejulio: I think that `byName` would be good, e.g.

``` java
class ByNameOperationBuilder
...
OperationBuilder.byName("redeploy")
```

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/cmd/ExecuteOperationCommand.java
ppalaga, line 59: > I still want to see in the log message that a request was received to execute operation O on resource R.

Hm... yes, the message we log in https://github.com/hawkular/hawkular-agent/blob/master/hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/cmd/AbstractResourcePathCommand.java#L76 does not contain the name of the Operation.

> Or maybe we could make these attributes not final to change the [Execute Operation] and [DMR Node] on runtime.

@josejulio , yeah, that would solve the problem, but I'd personally vote for keeping those final so that there is no state stored outside the `execute()` method. It is true, that we actually do not have such a contract here (whereas we have it in `command-gateway`), but immutability is still worth keeping because it makes the design cleaner and perhaps also because it allows JIT compiler to do more aggressive optimalizations. I see two options how to keep the GenericCommand and its parent immutable and get more informative log and error messages:

(a) add a `getOperationName` method to `AbstractResourcePathCommand` like this

```
    protected String getOperationName(BasicMessageWithExtraData<REQ> envelope) {
        return operationName;
    }
```

and use the value it returns in error and log messages. 

In `ExecuteOperationCommand`, it could simply be overriden:

```
    protected String getOperationName(BasicMessageWithExtraData<ExecuteOperationRequest> envelope) {
        return envelope.getBasicMessage().getOperationName();
    }
```

(b) Do not make `ExecuteOperationCommand` a subclass of `AbstractResourcePathCommand`

I tend to prefer (a)

WDYT?

josejulio:
You are right, i didn't notice it until now. I'll fix it up to be in line.

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
pilhuhn: On 18 Oct 2015, at 22:04, josejulio wrote:

> I think that `byName` would be good, e.g.
> 
> ``` java
> class ByNameOperationBuilder
> ...
> OperationBuilder.byName("redeploy")
> ```

+1

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/cmd/ExecuteOperationCommand.java
josejulio, line 59: I prefer (a), seems to be a cleaner solution. Besides, ExecuteOperationCommand already benefits from being a subclass of AbstractResourcePathCommand.

jmazzitelli, line 59: I don't care where the logging happens, I just wanted to make sure that "somewhere" we log the fact that a request came in to execute operation named O on resource R. However best you make that happen (or if it already happens, then great!) is up to you.

josejulio:
Updated the PR with the proposed changes.

ppalaga:
Thanks, @josejulio !

josejulio:
You're welcome!

PR #93, opened 2015-10-21
Log the WebSocket comm

ppalaga:


ppalaga:

This are the first steps towards a model that can emit not only the new resources and types added to the inventory but also the deletions and updates. In these commits I basically tried to simplify the model so that it becomes easier to introduce the named functionality.

jmazzitelli:
is this going to be related to https://issues.jboss.org/browse/HWKAGENT-20

I guess I will stop looking into that

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/inventory/TypeSet.java
```diff
@@ -0,0 +1,94 @@
...
+    public Map<Name, T> getTypeMap() {
+        return typeMap;
+    }
+
+    public boolean isEnabled() {
+        return enabled;
+    }
+
+    public boolean isDisabledOrEmpty() {
+        return isEnabled() || typeMap == null || typeMap.isEmpty();
```
jmazzitelli: need to negate isEnabled in the return statement with ! operator:

   return !isEnabled() || ...

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/inventory/Operation.java
```diff
@@ -16,18 +16,12 @@
  */
 package org.hawkular.agent.monitor.inventory;
 
-public abstract class Operation<RT extends ResourceType<?, ?, ?, ?>> extends NamedObject {
+public abstract class Operation extends NamedObject {
 
-    private final RT resourceType;
     private String operationName;
 
-    public Operation(ID id, Name name, RT resourceType) {
```
jmazzitelli, line 24: why remove this? Its possible that given an operation, we'll want to know what resource type it was associated with. Without this link we lose that context.

same with ResourceConfigurationPropertyType

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/inventory/ResourceConfigurationPropertyType.java
```diff
@@ -16,16 +16,9 @@
  */
 package org.hawkular.agent.monitor.inventory;
 
-public abstract class ResourceConfigurationPropertyType<RT extends ResourceType<?, ?, ?, ?>> extends NamedObject {
+public abstract class ResourceConfigurationPropertyType extends NamedObject {
 
-    private final RT resourceType;
-
-    public ResourceConfigurationPropertyType(ID id, Name name, RT resourceType) {
```
jmazzitelli, line 23: why remove this? Its possible that given a res config prop type, we'll want to know what resource type it was associated with. Without this link we lose that context.

same with Operation

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/inventory/TypeSet.java
ppalaga: +1

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/inventory/Operation.java
ppalaga, line 24: I removed it mainly, because I have not seen that it would be used now. 

The second, general point is that `resourceType` inside `Operation` brings a circular reference chain between `ResourceType` and `Operation` which is mostly not a good idea. E.g. it is not possible to make both classes immutable.

Thirdly, I thought that we probably won't need the link to `ResourceType`  in `Operation` if we can do well without it in `MetricType` and `AvialType`.

The same holds for `ResourceConfigurationPropertyType`.

Anyway, if there comes a situation, where the link will be inevitable, I am ready to introduce it back. But I still vote for removing it now.

hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/inventory/ResourceConfigurationPropertyType.java
ppalaga, line 23: For the the same reasons as with `Operation` (see above).

ppalaga:
7847012 fixes the missing `!` in `TypeSet.isDisabledOrEmpty()`.

@jmazzitelli is this good to merge now?

PR #96, opened 2015-10-22
HAWKULAR-740 - Added a new itest-debug profile.

jpkrohling:


hawkular-agent-itest-parent/hawkular-agent-itest/pom.xml
```diff
@@ -306,5 +310,13 @@
         </plugins>
       </build>
     </profile>
+
+    <profile>
+      <id>itest-debug</id>
```
jmazzitelli: please make this itest.debug so it follows the same syntax as the one we already have "test.debug" - otherwise, people will complain they can't remember when they need . or - with itest or unit tests.

PR #101, opened 2015-11-02
Major refactoring  * reduce the number of classes by using generics where possib

ppalaga:


ppalaga:
@jmazzitelli sorry that this took so long and sorry, that I have no more time today to rebase and fix the remaining compilation issues. I'd prefer to put more comments into the code before you start a real review but your early comments are welcome too :)

ppalaga:
@jmazzitelli 1267d9f is the state that compiles and where both unit and itests are passing. I am going to write comments into the code now.

ppalaga:
fb5eabd brings better class and package names and some class movements among packages

jmazzitelli:
There's so many files, git hub won't show me the diff of MonitorService for me to comment on it. But bootStorageAdapter.getUrl() (which is the url attribute of storageAdapter) can be explicitly defined by the user in standalone.xml or not. It used to be "url" was the only way to specify the hawkular server URL, but then we introduced the ability to define the outbound socket binding. So now if you specify the outbound socket binding you don't have to specify URL. In this new code (lines 115 and following) it doesn't seem to take into account that URL might be specified by the user. We actually could take that out of standalone.xml (since really we want to support the "official" way WildFly subsystems should define output connections - and that's with outbound socket bindings, not with URLs) but I think our tests and other things still use the URL attribute. In any event, I think this code no longer works the way it did - what if the user specifies the URL attribute explicitly in standalone.xml - does that still work? (remember, tenantID is NOT specified in his use-case - that is only used when in METRICS only mode - in the standard hawkular mode, the url can be specified, but tenantID is not).

Separately, in the same file, if we are in HAWKULAR mode, you cannot override or set tenantID manually - if the user does, we need to log a warning, and we should override it with what accounts tells us is our tenant ID. If we fail to get a tenantID from accounts, we should error out like we did before or better, maybe we just keep retrying if we failed to connect to accounts (but if we did connect to accounts and just got authentication failure, we need to error out) - the hawkular credentials determines what our tenant ID when in HAWKULAR mode. If the hawkular server can't tell us what our tenant ID is, don't let the user try to hijack or use a tenant ID that isn't confirmed by hawkular server (this is lines 157-194).

ppalaga:
About 24572a4 :
- rebased on top of the current master 
- it brings JavaDoc 
- it brings a little bit better use of type params - e.g. `EndpointService` now has just three type params compared to five in the previous iterartion.

ppalaga:
> There's so many files, git hub won't show me the diff of MonitorService for me to comment on it. 

Hm... indeed. To work it around I can only think of commenting here, eventually with a link to a line from the given commit, like I do later in this post.

> But bootStorageAdapter.getUrl() (which is the url attribute of storageAdapter) can be explicitly defined by the user in standalone.xml or not. It used to be "url" was the only way to specify the hawkular server URL, but then we introduced the ability to define the outbound socket binding. So now if you specify the outbound socket binding you don't have to specify URL. In this new code (lines 115 and following) it doesn't seem to take into account that URL might be specified by the user. 
> 
> We actually could take that out of standalone.xml (since really we want to support the "official" way WildFly subsystems should define output connections - and that's with outbound socket bindings, not with URLs) but I think our tests and other things still use the URL attribute. In any event, I think this code no longer works the way it did - what if the user specifies the URL attribute explicitly in standalone.xml - does that still work? (remember, tenantID is NOT specified in his use-case - that is only used when in METRICS only mode - in the standard hawkular mode, the url can be specified, but tenantID is not).

Ah, a good catch! I think I have understood how I broke the old logic. The block starting at https://github.com/ppalaga/hawkular-agent/blob/24572a4f46c1e8c90c45fd64b6a5ddfec5d3be98/hawkular-wildfly-monitor/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java#L133
should be executed only if `bootStorageAdapter.getUrl()` returned `null`, right?

Fixed here https://github.com/ppalaga/hawkular-agent/commit/cbb4733465b818e0f12834e525a91b8c0405989a#diff-6bc5e02300bbc2e87957a09fae5dc676R133

> Separately, in the same file, if we are in HAWKULAR mode, you cannot override or set tenantID manually - if the user does, we need to log a warning, and we should override it with what accounts tells us is our tenant ID. If we fail to get a tenantID from accounts, we should error out like we did before or better, maybe we just keep retrying if we failed to connect to accounts (but if we did connect to accounts and just got authentication failure, we need to error out) - the hawkular credentials determines what our tenant ID when in HAWKULAR mode. If the hawkular server can't tell us what our tenant ID is, don't let the user try to hijack or use a tenant ID that isn't confirmed by hawkular server (this is lines 157-194).

Yes, added both the warning https://github.com/ppalaga/hawkular-agent/commit/cbb4733465b818e0f12834e525a91b8c0405989a#diff-6bc5e02300bbc2e87957a09fae5dc676R201 and the exit https://github.com/ppalaga/hawkular-agent/commit/cbb4733465b818e0f12834e525a91b8c0405989a#diff-6bc5e02300bbc2e87957a09fae5dc676R435

ppalaga:
@jmazzitelli 599b952 brings an untested aggregation for DMR. The aggregation for JMX and Platform is missing. I must break now.

ppalaga:
@jmazzitelli would you please have a look esp. at the last commit? I hope I did the aggregation logic right.

Otherwise this is ready to be merged from my PoV

PR #102, opened 2015-11-04
HAWKULAR-585 - Updated Accounts - Cassandra datastore

jpkrohling:


ppalaga:
Let's do this after https://github.com/hawkular/hawkular-agent/pull/101

ppalaga:
merged as this one only changes poms that should not conflict with #101

PR #104, opened 2015-11-06
Update agent and itests to inv 0.8.0-Final where feeds moved out of environments.

metlos:


jmazzitelli:
travis failed, but i don't care. I'm merging this. @ppalaga  found that the failures are due to the welcome message not getting sent first thing and the tests don't expect that. we will fix that in the future. for now, this is getting merged so I can release the agent

PR #105, opened 2015-11-07
rename modules and directories to start with hawkular-wildfly-agent

ppalaga:


ppalaga:
@jmazzitelli this is the second reverted commit. Would you please review and eventually merge?

PR #107, opened 2015-11-10
HWKAGENT-30 - fix scheduler

jmazzitelli:


jmazzitelli:
right now this just adds a bunch of javadoc. I'm trying to find out why the scheduler jobs aren't running

PR #108, opened 2015-11-10
Simplify WebSocket integration tests and make the debugging easier

ppalaga:


ppalaga:
@jmazzitelli I am not 100% sure if I have understood the purpose of the diagnostics right, but the javadoc says that the JDR/JMX meters and timers are there to measure the collection of metrics and avails and therefore I think that we should meausure only inside measureAvials/Metrics as I do in [wrong]f11af84[/wrong] [right] c4bbad1 [/right] : https://github.com/ppalaga/hawkular-agent/commit/c4bbad1c5bc48119df5e8ab79986e6966300515e#diff-9ebba7f1d6cca76b7efa1ada207f756fR254 WDYT?

In accordance with my assumption f11af84 introduces a generic `SamplingDiagnostics` that is the same for every protocol and hence we can both cover Platform with it and at the same time, Drivers do not need to know about the diagnostics at all. Which is both cleaner, I think.

[edit] To put it in other words: As it stands before c4bbad1 , the timer and meter are invoked also during the discovery, which is IMO not right and which c4bbad1 addresses [/edit]

ppalaga:
f5a909e and c4bbad1 failed only because the log got too long. f11af84
 just turns off trace for OperationBuilder during itests to solve that.

And not to forget about f5a909e: it is just a simplification of the itests. It improves the debugging much : https://travis-ci.org/hawkular/hawkular-agent/builds/90302837#L1479 and removes the mockito dependency altogether.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/protocol/EndpointService.java
```diff
@@ -245,7 +250,13 @@ public void measureAvails(Collection<MeasurementInstance<L, AvailType<L>>> insta
             Driver<L> driver = session.getDriver();
             for (MeasurementInstance<L, AvailType<L>> instance : instances) {
                 AttributeLocation<L> location = instance.getAttributeLocation();
-                Object o = driver.fetchAttribute(location);
+                Object o = null;
+                try (Context timerContext = diagnostics.getRequestTimer().time()) {
```
jmazzitelli, line 254: the problem here is you are not just timing the actual execute of the DMR invocation - you are timing everything around it. I ONLY want diagnostics when you make the DMR call - so limit the amount of extra work the timer is including. There is too much else being including in this timer. Just follow the code in fetchAttribute and see all the extra stuff we are timing.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/protocol/dmr/DMRDriver.java
```diff
@@ -126,40 +119,30 @@ public boolean attributeExists(AttributeLocation<DMRNodeLocation> location) {
...
-        ReadAttributeOperationBuilder<?> opBuilder = OperationBuilder
-                .readAttribute()
-                .address(location.getLocation().getPathAddress())
-                .resolveExpressions()
-                .name(useAttribute);
 
-        // time the execute separately - we want to time ONLY the execute call
-        OperationResult<?> opResult;
-        try (Context timerContext = diagnostics.getDMRRequestTimer().time()) {
-            opResult = opBuilder.execute(client);
```
jmazzitelli, line 138: this - this is what I want to time. and just that invocation I want diagnostics on. I want to know the performance of the underlying DMR request.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/protocol/EndpointService.java
```diff
@@ -282,7 +293,13 @@ public void measureMetrics(Collection<MeasurementInstance<L, MetricType<L>>> ins
             Driver<L> driver = session.getDriver();
             for (MeasurementInstance<L, MetricType<L>> instance : instances) {
                 AttributeLocation<L> location = instance.getAttributeLocation();
-                Object o = driver.fetchAttribute(location);
+                Object o = null;
+                try (Context timerContext = diagnostics.getRequestTimer().time()) {
+                    o = driver.fetchAttribute(location);
+                } catch (Exception e) {
+                    diagnostics.getErrorRate().mark(1);
```
jmazzitelli, line 300: and this is wrong - if we get an exception here, this isn't necessarily from a problem with the DMR connection - it could have been an exception from else where. Again, the error rate for the DMR diagnostics is ONLY to capture errors we had in the DMR communications, nothing extra inside of the agent should trigger a DMR error rate increment.

ppalaga:
deb221f puts the diagnostics measurements back to protocol-specific Drivers, newly including Platform

PR #110, opened 2015-11-11
HWKAGENT-32 Cannot collect platform resources due to a NPE in PlatformLocationRe

ppalaga:


ppalaga:
@jmazzitelli this applies the same logic in `PlatformLocationResolver` we already have in `DMRLocationResolver` thus allowing for `null` base paths and basically interpreting them as root. 

I was not able to reproduce the issue quickly enough and therefore I am also not able to validate the fix. But given that we already have the same logic in place in DMR, there is a good chance that the fix will work. I must turn to other tasks now.

PR #113, opened 2015-11-17
Updated to Accounts 1.1.1.Final

jpkrohling:


jpkrohling:
Will re-submit with new version.

pilhuhn:
@jpkrohling Is that still relevant? Especially given that accounts is at 2.0.6+ now

jpkrohling:
The 2.x branch is for WF 10, so, if there's a subsequent commit here that is needed in the main Hawkular, it would need to wait for WF 10 to arrive there before consuming this new version of the agent. 

So, 1.x is still appropriate.

jpkrohling:
By the way, it seems this has already somehow arrived at master. So, this can be closed.

PR #116, opened 2015-11-18
Test correction / running on windows issues

sekigor:
Windows related issues running the tests:
Line spacing issue on ModulesTest
UnsupportedOperationException on some calls from oshi.hardware.Processor


ppalaga:
@sekigor thanks for the pull request. 

Could you please squash 6ece4fe and c11a138 into one commit and find a better (grammar and meaning) commit message ?  - e.g. you could use something like "ModulesTest not passing on Windows due to mismatching EOLs". 

In 6925836, you are fixing a different issue, and thus the commit message needs to be different. "Fix OshiPlatformCacheTest on Windows" or similar.

I'll add a couple of inline comments.

pilhuhn:
This is right now failing in checkstyle with lines in `OshiPlatformCahceTest` being too long. I can imagine that this needs another rebase against master to fix that.

PR #126, opened 2015-12-15
Upgrade to Accounts 1.1.4.Final, Command Gateway 0.10.5.Final, Commons 0.2.5.Fin

ppalaga:


ppalaga:
In 5e26a04 the tests are not passing

ppalaga:
7595524 should fix the itests. The original idea comes from @metlos (thanks!).

PR #127, opened 2015-12-16
Use the Accounts Feature Pack to run the integration tests

ppalaga:


ppalaga:
5a930e7 does not work

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceConfigurationBuilder.java
```diff
@@ -754,6 +754,7 @@ private static GlobalConfiguration determineGlobalConfig(ModelNode config, Opera
             throws OperationFailedException {
         boolean subsystemEnabled = getBoolean(config, context, SubsystemAttributes.ENABLED);
         String apiJndi = getString(config, context, SubsystemAttributes.API_JNDI);
+        System.out.println("apiJndi = "+ apiJndi);
```
jmazzitelli: remove this System.out.println before merging

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/SubsystemAdd.java
```diff
@@ -74,12 +69,12 @@ private void createService(final ServiceTarget target, final List<ServiceControl
         // create the builder that will be responsible for preparing the service deployment
         ServiceBuilder<MonitorService> svcBuilder;
         svcBuilder = target.addService(SubsystemExtension.SERVICE_NAME, service);
-        svcBuilder.addListener(verificationHandler);
         svcBuilder.setInitialMode(ServiceController.Mode.ACTIVE);
         service.addDependencies(svcBuilder);
 
         // bind the API to JNDI so other apps can use it, and prepare to build the binder service
         String jndiName = configuration.getApiJndi();
+        System.out.println("About to use jndiName = "+ jndiName);
```
jmazzitelli: remove this System.out.println before merging

hawkular-wildfly-agent/src/test/resources/org/hawkular/agent/monitor/extension/subsystem.xml
```diff
@@ -27,18 +27,20 @@
...
   <diagnostics enabled="false"
-               reportTo="LOG"
                interval="1"
                timeUnits="minutes"/>
 
-  <storage-adapter type="HAWKULAR"
-                   username="jdoe"
+  <!-- default values omitted:
+          type="HAWKULAR"
+          inventoryContext="/hawkular/inventory/"
```
jmazzitelli: we shouldn't even have to mention inventoryContext or metricsContext even in comments - these should never change and no one should really care these attributes exist. These are here only for the future in case we add the ability for the user to be able to change kettle's context from /hawkular to something else. So take those two out even in the comments.

hawkular-wildfly-agent/src/test/resources/org/hawkular/agent/monitor/extension/subsystem.xml
```diff
@@ -16,9 +16,9 @@
     limitations under the License.
 
 -->
+<!-- default value omitted: apiJndiName="java:global/hawkular/agent/api" -->
```
jmazzitelli: don't even worry about commenting this here. The user will rarely, if ever, change that default attribute value - in fact, they probably don't even need to know it exists.

hawkular-wildfly-agent/src/main/resources/subsystem-templates/hawkular-wildfly-agent.xml
```diff
@@ -0,0 +1,94 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
jmazzitelli, line 1: What is this file for? The agent shouldn't have to care about messaging-activemq existing, or actually anything in here that I can see.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/SubsystemAdd.java
```diff
@@ -101,13 +96,11 @@ private void createService(final ServiceTarget target, final List<ServiceControl
             svcBuilder.addDependency(binderServiceName);
 
             // install the binder service
-            ServiceController<?> binderController = binderBuilder.install();
-            newControllers.add(binderController);
+            binderBuilder.install();
```
jmazzitelli, line 98: I assume this still magically works?

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/SubsystemAdd.java
```diff
@@ -101,13 +96,11 @@ private void createService(final ServiceTarget target, final List<ServiceControl
...
         }
 
         // install the monitor service
-        ServiceController<MonitorService> svcController = svcBuilder.install();
-        newControllers.add(svcController);
+        svcBuilder.install();
```
jmazzitelli, line 102: Still works even though svcController is not being added?

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceConfigurationBuilder.java
ppalaga: +1

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/SubsystemAdd.java
ppalaga: +1

hawkular-wildfly-agent/src/test/resources/org/hawkular/agent/monitor/extension/subsystem.xml
ppalaga: OK, I'll remove the comments. I was not sure you'll ack this strategy of removing defaults.

hawkular-wildfly-agent/src/main/resources/subsystem-templates/hawkular-wildfly-agent.xml
ppalaga, line 1: This is where the WF build plugins (feature pack and provisioning) expect the fragment with the agent subsystem's configuration. It should not be empty, I just forgot about it when I was asking your what the default config should be. The content from `hawkular-wildfly-agent-wf-extension/src/main/resources/system/layers/base/org/hawkular/agent/main/subsystem-snippet.xml` should basically be moved here and the Installer should be somehow adapted. I will polish this part later.

The objective of the present PR is to get the Agent working on WF10, Artemis and all the changed upstream components.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/SubsystemAdd.java
ppalaga, line 98: Yes, this is exactly what @ctomc has done on many places in WF too.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/SubsystemAdd.java
ppalaga, line 102: If we'll find this not working we will blame @ctomc ;)

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/SubsystemAdd.java
ctomc, line 98: yes, this is possible since WildFly 8.x times IIRC.
We also overloaded the performRuntime method and depreacted old one that had list of newControllers in it.
It is all done behind the scenes now.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/SubsystemAdd.java
ctomc, line 102: you can do that :)

ppalaga:
345f83d is still not working. Looks like all cmdgw releated issues are solved, but the inventory is not working properly. 345f83d takes the inventory srcdep 4af1c21 that is available in https://github.com/hawkular/hawkular-inventory/pull/208
@metlos could you please have a look?

ppalaga:
The itests are passing with 1a43064

ppalaga:
It can be seen in the log that Agent cannot store metrics:

```
ERROR [org.hawkular.agent.monitor.storage.MetricsOnlyStorageAdapter] (OkHttp http://127.0.0.1:8080/hawkular/metrics/metrics/data) HAWKMONITOR010010: Failed to store metric data: {"counters":[],"availabilities":[],"gauges":[{"data":[{"value":0.0,"timestamp":1450386786893}],"id":"MI~R~[01e04c14-9b97-48c8-afbc-993d3c149308/Local~/deployment=hawkular-inventory-dist.war/subsystem=ejb3/stateless-session-bean=OrganizationMembershipServiceImpl]~MT~Stateless Session EJB Metrics~Pool Current Size"}]}: java.io.IOException: status-code=[503], reason=[Service Unavailable], url=[http://127.0.0.1:8080/hawkular/metrics/metrics/data]
    at org.hawkular.agent.monitor.storage.MetricsOnlyStorageAdapter$1.onResponse(MetricsOnlyStorageAdapter.java:118)
    at com.squareup.okhttp.Call$AsyncCall.execute(Call.java:170)
    at com.squareup.okhttp.internal.NamedRunnable.run(NamedRunnable.java:33)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
```

https://travis-ci.org/hawkular/hawkular-agent/builds/97543989#L2334

@jsanda do you happen to have an idea what could be wrong?

This is using metrics 0.10.0.Final - is that maybe a problem?

edit: @jsanda as @jmazzitelli correctly points out, the above 503s are during shutdown. But there are 401s https://travis-ci.org/hawkular/hawkular-agent/builds/97543989#L2582

ppalaga:
@jsanda , @jmazzitelli I do not see those 401s locally. FWIW, i added `hawkular.metrics.waitForService` in #129  https://github.com/ppalaga/hawkular-agent/commit/d0a39f9165842a1787dd93ca98782fca24e16c05

PR #129, opened 2015-12-18
Tidy up poms, esp. remove unneeded parts and manage in top level pom

ppalaga:


ppalaga:

PR #136, opened 2016-01-07
Fix parent artifactId in hawkular-wildfly-agent-example-jndi

ppalaga:


ppalaga:

PR #137, opened 2016-01-12
HWKAGENT-47 Generate subsystem-snippet.xml out of subsystem-templates/hawkular-wildfly-agent

ppalaga:


ppalaga:
@mazz this aims at having a single subsystem configuration xml fragment. It is located in [hawkular-wildfly-agent/src/main/resources/subsystem-templates/hawkular-wildfly-agent.xml](https://github.com/ppalaga/hawkular-agent/commit/ac5462eaaf0255ba912fe3439052068ecd805a1e#diff-e89482f1be283412d2125c5e3ff45399L1) where the feature-pack plugin expects it. The `subsystem-snippet.xml` is now generated from `hawkular-wildfly-agent.xml` in https://github.com/hawkular/hawkular-agent/commit/a752af743ac030101258773bf68cd048636ad63c#diff-534639d72b5358e7cb469ddec69bd254R112

ppalaga:
@mazz could you please review and eventually merge?

mazz:
Wrong mazz

jmazzitelli:
@ppalaga  I assigned you to https://issues.jboss.org/browse/HWKAGENT-47 since that is the JIRA covering this work

ppalaga:
@mazz I am sorry and will do my best to notice :)

mazz:
No problem. I just wanted to make sure you weren't under the false impression that your intended recipient got their message

hawkular-wildfly-agent-feature-pack/src/main/resources/modules/system/layers/base/org/hawkular/agent/main/module.xml
```diff
@@ -83,6 +83,7 @@
     <module name="org.jboss.vfs"/>
 
     <!-- other dependencies we need -->
+    <module name="javaee.api"/> <!-- bus-commons -->
```
jmazzitelli, line 86: Does bus-commons really need JEE? I think that would be bad because it means we won't be able to install the agent in any WildFly-core server or any server that doesn't have JEE (like a trimmed down swarm instance).

hawkular-wildfly-agent-feature-pack/src/main/resources/configuration/standalone/template.xml
```diff
@@ -18,4 +18,85 @@
 
 -->
 <server xmlns="urn:jboss:domain:4.0">
+
+  <extensions>
+        <?EXTENSIONS?>
+  </extensions>
+
+  <management>
```
jmazzitelli, line 26: I don't know what this file is - does this mean the agent feature pack requires these management realms? The agent itself shouldn't need these management realms

hawkular-wildfly-agent-feature-pack/src/main/resources/modules/system/layers/base/org/hawkular/agent/main/module.xml
ppalaga, line 86: `javaee.api` provides the JMS 2.0 API. Let me have a look if this could be done differently.

jmazzitelli, line 86: This is another reason why we can't have BasicMessage and related classes importing JMS classes - because there will be environments where that won't be available. The agent's use of bus-common is only BasicMessage and the JSON stuff. It doesn't even use the bus consumer/producer stuff.

hawkular-wildfly-agent-feature-pack/src/main/resources/configuration/standalone/template.xml
ppalaga, line 26: `subsystems.xml` and `template.xml` are there to produce all standalone*.xml files for a runnable WF distro.

I admit I have just mechanically taken the files from WildFly feature pack [1] and enhanced them with the Agent stuff. I assumed my mechanical approach is well justified by the fact that the Agent feature pack depends on the WF feature pack [2]. You can clearly see from the blame [2] that it has been for a long time like that already. In other words, given that our feature pack depends on WF feature pack, nobody (incl. me) should assume that the agent will be able to run on top of WF Core [3] or some stripped-down version of WF.

If you believe the Agent feature pack should depend on something than the WF feature pack, please state it clearly. But in such a case, I'd vote for merging the present, WF feature pack based PR first and try to change the base after that.

[1] https://github.com/wildfly/wildfly/tree/master/feature-pack/src/main/resources/configuration/standalone
[2] https://github.com/ppalaga/hawkular-agent/blame/ac5462eaaf0255ba912fe3439052068ecd805a1e/hawkular-wildfly-agent-feature-pack/feature-pack-build.xml#L23
[3] https://github.com/wildfly/wildfly-core/tree/master/core-feature-pack

hawkular-wildfly-agent-feature-pack/src/main/resources/modules/system/layers/base/org/hawkular/agent/main/module.xml
ppalaga, line 86: > Let me have a look if this could be done differently.

Given my previous statement https://github.com/hawkular/hawkular-agent/pull/137/files#diff-05665f6a96889fd9eb46439a463b6646R26 as long as this feature pack depends on wildfly-feature-pack, the  `javaee.api` module is guaranteed to be there and I do not need to look for another solution.

ppalaga, line 86: > The agent's use of bus-common is only BasicMessage

Agent really does not use JMS? - Actually, you may well be right :)

In such a case, I'd say that the proper solution should consist in splitting `BasicMessage` & Co. from bus and putting it in a new artifact, call it e.g. `commons-messages`.

As an independent but related topic, we should have a look at the possibility to move message classes from `cmdgw` to the new `commons-messages`. That would allow Alerts to drop the dependency on `cmdgw`.

jmazzitelli, line 86: That might not be a bad idea. Split out BasicMessage and its related classes, with all the JSON processing that goes with it, into a smaller commons library.

Agent does NOT use JMS. Never did. The only time it interfaced with the bus was many releases ago when it went through the REST api that sat in front of ActiveMQ. But that was solely a temporary solution and has gone away now. But it never used JMS classes directly.

hawkular-wildfly-agent-feature-pack/src/main/resources/configuration/standalone/template.xml
jmazzitelli, line 26: I have not attempted to run on Wildfly-Core. I don't even know where that project is at the moment (is that still going to be released now that we have swarm?). But back in the day, the talk was we want hawkular agent to be small enough and not have large amounts of dependencies so it can run inside of WildFly-Core. This was the initial thinking of Heiko Braun when he wrote the first agent prototype - whether it is something we still want to be able to support is something for @theute and @pilhuhn to confirm. I always assumed we would eventually want to install agents inside of WildFly-Core and swarm-based containers that strip out JEE stuff. I wonder if @bobmcwhirter has tried installing agents in smaller swarm containers?

bobmcwhirter, line 26: I think we’ve used it on JAX-RS Swarm containers that only include JAX-RS, Undertow, and whatever it absolutely requires.

I haven’t noticed it having too many onerous requirements.  But I could also be forgetful.

-Bob

ppalaga, line 26: @bobmcwhirter do these `subsystems.xml` and `template.xml` files have any significance for WF Swarm?

bobmcwhirter, line 26: Currently they don’t.  We read them as humans to translate to our Java API to configure it.  So they’re useful to me, as a person.

That said, @heiko-braun is doing work to let -swarm play with standalone.xml and such, so we may soon reach the point where we are consuming or otherwise inspecting the XML fragments you guys ship in the feature-packs.

-Bob

ppalaga, line 26: @bobmcwhirter thanks for the info.

@jmazzitelli , as I said, I vote for merging this as is now and experiment with wf-core later, when we have split the message classes to a new JMS independent maven module.

ppalaga:
The new set of changes ending with 9579f98
- Solves the problem with the bad parent version 0.15.3.Final-SNAPSHOT in some itest module
- Fixes the dependencies of the itest module to be the hawkular-wildfly-agent-itest-feature-pack rather than hawkular-wildfly-agent-feature-pack + accounts-fp
- Upgrades to parent 33 and related component upgrades. Note that metrics is still at 0.10.0.Final, I'll start working on an upgarde to 0.11.0.Final now.

@jmazzitelli could you please re-review and either state what is still missing or merge?

ppalaga:
5a55e0b adds an upgrade to Metrics 0.11.0.Final. It was more flawless than I expected.

jmazzitelli:
Do not merge this. Even though it builds and itests pass, if you then try to build hawkular with this agent, the hawkular server aborts at boot time with a bad configuration. I think @ppalaga  is looking into that. Once that is fixed, we'll make sure this PR works with it and when that is confirmed, we can merge.

ppalaga:
@jmazzitelli this PR shows what needs to be done in Hawkular: https://github.com/hawkular/hawkular/pull/781

PR #142, opened 2016-01-25
hwkagent 52 - don't deploy if operation doesn't exist

jmazzitelli:


tsegismont:
The changes look good to me. I just wonder why the "user visible name" is used in `ExecuteOperationCommand`. Doesn't that prevent to change display names at will?

jmazzitelli:
> I just wonder why the "user visible name" is used in ExecuteOperationCommand.
> Doesn't that prevent to change display names at will?

Good point. The operation name comes from the metadata found in standalone.xml. For example:

&lt;operation-dmr name="JDR" operationName="generate-jdr-report" path="/subsystem=jdr" />

The client will tell the server to execute the "JDR" operation, not the actual DMR operation name of "generate-jdr-report".

So when I said, "user-visible", that probably was not a good way to say it. I was trying to avoid saying, "We want to use the name of the operation, not the operationName" :-)

We could i18n this "name", for example, in the future. But that i18n phrase would not be what the client is sending. The client is sending the operation's "name" not the "operationName"

The word "name" is overloaded here, so its hard to describe :) But my thinking was the client will know about the operation metadata (both name and operationName) but it shouldn't care about "operationName" because that is solely for the use of the agent when sending the execution request to the actual managed resource.

ppalaga:
@jmazzitelli is this not a nice occasion to add an itest for `DeployApplicationCommand`?

jmazzitelli:
Yes, I should be adding a test for this. I noticed we had hello world war in the itest that I thought we were already deploying, but I notice we aren't deploying it in an itest and thus not really testing the deploy operation.

jmazzitelli:
I'm going to build a deploy itest via a separate JIRA. See: https://issues.jboss.org/browse/HWKAGENT-54

ppalaga:
@jmazzitelli thanks, basically +1 for the state ending with  bcdda6b . I vote for squashing the commits together and merging afterwards.

PR #146, opened 2016-01-28
HWKAGENT-53 Agent should reuse TestWebSocketClient and EchoCommandITest from Command Gateway

rubenvp8510:
I'm using dependenciesToScan for run cmdgw tests,but that involves to run all tests on cmdgw-itest, it should be a better way to do it for run only the echo test.


jmazzitelli:
Take a look at https://github.com/hawkular/hawkular-agent/commit/79ffbadef203ecd38ad728a62911e5c5047ea377

I already renamed the package from cmdgw -> agent and moved files appropriately. So do a pull and rebase and you'll pick up those changes already.

ppalaga:
@rubenvp8510 yes, the EchoCommandITest is executed, which great, but https://issues.jboss.org/browse/HWKAGENT-53 also requires to remove `TestWebSocketClient` from Agent's itests and make the Agent's itests use the class from command gateway. Clould you please fix that too?

rubenvp8510:
@ppalaga  Yes, I can do that too.

rubenvp8510:
@ppalaga  I think I finished, now `TestWebSocketClient` is on cmdgw, and the agent tests uses it.

ppalaga:
@rubenvp8510 looks good, thanks. If you could please prepend the commit message with `HWKAGENT-53` so that Jira can auto-close. I'll merge after that.

PR #150, opened 2016-02-05
HWKAGENT-57 set avails to DOWN when agent is going down

jmazzitelli:


jshaughn:
It looks like you can optionally set to UNKNOWN as opposed to DOWN, which is a good thing because an agent being down does not necessarily mean the server, or it's components, are down.

jshaughn:
given the prior comment, LGTM

PR #151, opened 2016-02-09
HWKAGENT-58 installer should be able to configure agent for metrics-only

jmazzitelli:


.travis.yml
```diff
@@ -17,7 +17,7 @@ notifications:
 install:
 - mvn -version -B
 script:
-- mvn -s .travis.maven.settings.xml verify -Pitest -Dfailsafe.useFile=false | grep -vF "[INFO] Downloading:" | grep
+- mvn -s .travis.maven.settings.xml verify -Pitest -Dfailsafe.useFile=false -Dlicense.skip=true | grep -vF "[INFO] Downloading:" | grep
```
jpkrohling, line 20: I understand the frustration with the license check, but I don't think it's a good idea to skip it. We better suffer a bit more and fix whatever there is to fix there, instead of just ignoring this.

hawkular-wildfly-agent-installer/src/main/java/org/hawkular/wildfly/agent/installer/AgentInstaller.java
```diff
@@ -349,8 +352,21 @@ private static String createSecurityRealm(String keystoreFile, String keystorePa
...
+            xml.append(" type=\"METRICS\"");
+            if (tenantId == null || tenantId.isEmpty()) {
+                throw new IllegalArgumentException("You must specify tenant-id when in metrics-only mode");
+            }
+        } else {
+            xml.append(" type=\"HAWKULAR\"");
+        }
+
+        if (tenantId != null && !tenantId.isEmpty()) {
+            xml.append(" tenantId=\"" + tenantId + "\"");
```
jpkrohling, line 368: From what I'm understanding on this code, it allows to include a tenantId on a non-metrics type as well. I think this should _not_ be allowed, as it breaks the user's expectations: this is either ignored later on, or is used and the backend rejects the request.

hawkular-wildfly-agent-installer/src/main/java/org/hawkular/wildfly/agent/installer/InstallerConfiguration.java
```diff
@@ -135,6 +137,20 @@ static Options buildCommandLineOptions() {
                 .desc("The feed ID that the agent will use to identify its data.")
                 .numberOfArgs(1)
                 .build());
+        options.addOption(Option.builder()
+                .argName(InstallerConfiguration.OPTION_TENANT_ID)
+                .longOpt(InstallerConfiguration.OPTION_TENANT_ID)
+                .desc("The tenant ID that the agent will ask to be used. Usually only used when in metrics-only mode.")
```
jpkrohling, line 143: s/Usually only used/Should be only used/

jpkrohling:
First round done.

.travis.yml
jmazzitelli, line 20: This is only temporary. When we fix the license checker, this is going to be removed

hawkular-wildfly-agent-installer/src/main/java/org/hawkular/wildfly/agent/installer/AgentInstaller.java
jmazzitelli, line 368: yes, this allows a tenant ID to be specified for HAWKULAR mode. What this allows is when the agent starts up, IFF the tenant ID matches that of the agent's credentials, the agent starts up fine. If the tenantId configured here does NOT match the user's credentials, the agent will log an error and not start. So it works as a sort of "check" - kind of like allowing the person installing the agent to say, "I expect the agent to be associated with tenant ID "foo" - if the credentials given to the agent do not give it tenant Id "foo" the agent should not start."

We could throw an exception here if we don't like that "feature" - but, I didn't see any harm in allowing this check-tenant feature so I left it here.

hawkular-wildfly-agent-installer/src/main/java/org/hawkular/wildfly/agent/installer/InstallerConfiguration.java
jmazzitelli, line 143: unless we leave the "check-tenant" feature alone, as I describe above.

hawkular-wildfly-agent-installer/src/main/java/org/hawkular/wildfly/agent/installer/AgentInstaller.java
jpkrohling, line 368: I see no problem with that, if it acts the way you said. My only concern would be that the agent could send bad data to the backend, but it doesn't seem to be the case then.

jpkrohling:
Looks good to me.

.travis.yml
jpkrohling, line 20: OK, can you then record this in a JIRA? We should not forget to enable it back.

jmazzitelli, line 20: HWKAGENT-59

hawkular-wildfly-agent-installer/src/main/java/org/hawkular/wildfly/agent/installer/AgentInstaller.java
jmazzitelli, line 368: Now that I look at the code, the agent only logs a warning but it doesn't halt the agent startup. I don't like that :) There should be an exception thrown here and the agent startup should fail because the agent is starting with a tenant ID that doesn't match the explicit configuration in tenantID. I will fix it so the agent aborts startup after it logs this message.

``` java
if (useTenantId != null && !tenantIdFromAccounts.equals(useTenantId)) {
    log.warnIgnoringTenantIdFromXml(useTenantId, tenantIdFromAccounts);
}
```

PR #152, opened 2016-02-09
HAWKULAR-790 - be able to ask for the current state of all stat enabled flags

jmazzitelli:
be able to ask for the current state of all stat enabled flags.

even if you set a flag, the response will always include the known settings of all the flags.

using cmdgw snapshot because I can't use srcdep yet until that PR is peer-reviewed and merged first.
one that happens, we can adjust the pom here.


jmazzitelli:
one pass of the tests worked, the other didn't. I hate the travis setup. This is working. No idea why travis failed that second time around.

jshaughn:
I see the srcdep is updated so LGTM...

PR #153, opened 2016-02-11
Updated Accounts to 2.0.21.Final

jpkrohling:


ppalaga:
@jpkrohling @jmazzitelli Although, I tried some debugging yesterday, I do not have any quick fix for the license header issue. I propose to disable the lic. plugin for now and I'll fix it later after I am finished with tasks blocking M10.

jmazzitelli:
@jpkrohling if you can put back the -Dlicense.skip=true in this PR, once it goes green I'll merge.

Note that even though it will be disabled in travis, "normal" local builds will still run the license checker. So at least license check is still going to get done during local builds (I never turn it off locally - seems to work fine outside of travis).

jpkrohling:
Added.

jmazzitelli:
getting a "real" failure now: [ERROR] Failed to execute goal org.wildfly.build:wildfly-server-provisioning-maven-plugin:1.1.2.Final:build (server-provisioning) on project hawkular-wildfly-agent-itest: Execution server-provisioning of goal org.wildfly.build:wildfly-server-provisioning-maven-plugin:1.1.2.Final:build failed: java.lang.RuntimeException: java.lang.IllegalStateException: Could not resolve '[subsystem=hawkular-accounts-ejb3.xml, supplement=null] 

This is from the itests - so to replicate, you will probably need to pass in -Pitest when you build locally.

jpkrohling:
OK, that's a common issue for consumers of the feature pack. I wonder how I didn't get this before. Anyway, I'll update the PR with the fix.

jpkrohling:
Updated :)

PR #154, opened 2016-02-11
Add --debug to Travis build

ppalaga:


ppalaga:
This PR served only for debugging. The real fix is in https://github.com/hawkular/hawkular-agent/pull/155 thus closing this one.

PR #157, opened 2016-02-15
Upgrade to Command Gateway 0.10.11.Final Commons 0.3.5.Final Inventory 0.12.2.Fi

ppalaga:


ppalaga:

jmazzitelli:
I don't think this is right. Unless they changed their release numbers, I need the latest inventory and metrics (0.13 for both).

jmazzitelli:
We need to make sure the agent works when downgrading inventory and metrics like this. I put "Do Not Merge Yet" label on this PR until we are sure this is OK.

jpkrohling:
Yes, they should work, as it wasn't a "real" downgrading. For metrics, the required commits were cherry-picked from master into the branch. For inventory, a release of master was made under the 0.12.1 version. Perhaps @metlos can confirm the inventory's side, but I can confirm that the metrics version specified there is the one that we should be using for Hawkular Alpha 10.

PR #158, opened 2016-03-03
HWKAGENT-62 put agent subsystem in WF add-ons location

ppalaga:


ppalaga:

ppalaga:
ce2455d makes Checkstyle happy

jmazzitelli:
```
- configurationBldr.modulesHome("modules");
+ configurationBldr.modulesHome("modules/system/add-ons/hawkular-agent");
```

I am glad this looks to have been a simple one-line fix. I was crossing my fingers it didn't require more than that. :)

I assume you ran the installer to see that it went in the proper place and then ran the actual WF instance to see that the agent really does deploy and start up properly when placed in this add-ons directory?

ppalaga:
> I am glad this looks to have been a simple one-line fix

I also changed the location of the module.xml in the source tree.

> I assume you ran the installer

No, I have not. Don't we have an itest checking this?

jmazzitelli:
> No, I have not. Don't we have an itest checking this?

Even if we have an itest, its always a good idea to always just run the software. As for this, I don't know if an itest actually runs the installer and then runs the WF instance with that installed agent. What WILL test this is the build of hawkular distro - because the hawkular distro build uses the installer to install the agent. So if you build hawkular distro and run it, if you see the agent starting up, this works.

ppalaga:
OK, do not merge yet, I'll add such a test. I should actually also add a simple test checking that the Agent is writing to Metrics, because I suspect that the writes are not successful ATM, but the build is passing anyway.

ppalaga:
911f952 brings some integration tests to ensure that Agent installer is able to:
- Install on top of a plain WildFly
- that the WF with the freshly installed agent boots and writes data to Inventory and Metrics hosted on an external Hawkular instance

ATM, the named tests live in a separate Maven module `hawkular-wildfly-agent-installer-itest`. I started like that to be able to work in isolation from the tests available at present (to exclude interferences) but I am not particularly happy with the result, because there are as many as three container instances stared during the itests: once plain WF with Agent installed and two Hawkular instance. I think that it would be better to merge `hawkular-wildfly-agent-installer-itest` with the existing `hawkular-wildfly-agent-itest` so that there is just the plain WF and one Hawkular. However, if Travis is able to swallow the setup like this, I vote for merging 911f952 and eventually improve later so that this does not block Hawkular M11. WDYT, @jmazzitelli ?

jmazzitelli:
I totally forget the agent extension jar is packaged with the module subdirectories. So no code changes needed to be made, I guess - just package up the assembly so the jar has the 'add-ons' and that's all that was needed. Since travis is green, I'm good to merge this.

ppalaga:
> I'm good to merge this.

Could me or you release the Agent for M11 after merging?

jmazzitelli:
@ppalaga if you want to merge/release now, go ahead. I have nothing that needs to go into M11, so it can be done now or tomorrow, it doesn't matter. I can release if you can't, but I would have to do it tomorrow.

PR #159, opened 2016-03-04
initial ansible playbooks/roles

jmazzitelli:
@pilhuhn here's a PR with an initial set of Ansible roles/playbooks. See README.adoc for quick-start docs. I will blog about this, but first wanted to get your take.

Note that the Wildfly roles are really here just to make things easy on the person testing these. It does a download/install of WF 10 (so we can install an agent into it) and start/stop/status of the WildFly instances. The role we really care about is "fetch-and-install-hawkular-wildfly-agent" since that does the download and installation of the agent.

### DIFF
--- a/hawkular-wildfly-agent-ansible-playbook/roles/fetch-and-install-hawkular-wildfly-agent/tasks/main.yml
+++ b/hawkular-wildfly-agent-ansible-playbook/roles/fetch-and-install-hawkular-wildfly-agent/tasks/main.yml
@@ -1,0 +1,65 @@
+#
+# Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
+# and other contributors as indicated by the @author tags.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+---
+- name: Create Hawkular download tmp directory
+  file: path="{{hawkular_download_tmpdir}}" state=directory
+
+- name: Download Hawkular WildFly Agent installer
+
+  # Can't use due to bug: https://github.com/ansible/ansible-modules-core/issues/2166
+  #uri:
+  #  url="{{hawkular_server_url}}/hawkular/wildfly-agent/installer"
+  #  dest="{{hawkular_download_tmpdir}}/hawkular-wildfly-agent-installer.jar"
+  #  user="{{hawkular_server_url_username}}"
+  #  password="{{hawkular_server_url_password}}"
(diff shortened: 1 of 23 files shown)


jmazzitelli:
Note that each file blows up because of the license header stuff. I'll add those once I find out what a yaml header file looks like :)

hawkular-wildfly-agent-ansible-playbook/roles/fetch-and-install-hawkular-wildfly-agent/tasks/main.yml
```diff
@@ -0,0 +1,63 @@
...
+  #  user="{{hawkular_server_url_username}}"
+  #  password="{{hawkular_server_url_password}}"
+  #  force_basic_auth=true
+  #  method=POST
+  #  body=""
+  #  follow_redirects=all
+  #  validate_certs=no
+
+  shell:
+    "curl --silent --show-error --create-dirs --user '{{hawkular_server_url_username}}:{{hawkular_server_url_password}}' --output '{{hawkular_download_tmpdir}}/hawkular-wildfly-agent-installer.jar' {{hawkular_server_url}}/hawkular/wildfly-agent/installer"
```
pilhuhn, line 38: Can we guarantee that curl is installed?

hawkular-wildfly-agent-ansible-playbook/roles/fetch-and-unpack-wildfly/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
...
+# limitations under the License.
+#
+
+# This only works with Ansible 2.0 (because unarchive src is an http URL)
+---
+  - name: Create WildFly installation directory
+    file: path="{{wildfly_install_dir}}" state=directory mode=0755
+
+  - name: Download and unpack WildFly app server
+    unarchive:
```
pilhuhn, line 24: This always pulls from the remote repo no matter on how many hosts we install.
I had a playbook for pulling that into the ansible "server" and then use that url. See https://github.com/pilhuhn/hawkular-agent-ansible/blob/master/roles/fetch_wildfly/tasks/main.yml

hawkular-wildfly-agent-ansible-playbook/roles/fetch-and-install-hawkular-wildfly-agent/tasks/main.yml
jmazzitelli, line 38: We could use ansible to install curl :) but really, this was just a stop-gap for Ansible 1.9. Moving to Ansible 2.0 and I would have been able to use the uri: module instead (I wanted to use uri: module now, but I'm on Ansible 1.9 (Fedora 23 doesn't have 2.0 in its dnf repos) and that bug I mentioned in the comments breaks the uri: module).

hawkular-wildfly-agent-ansible-playbook/roles/fetch-and-unpack-wildfly/tasks/main.yml
jmazzitelli, line 24: This task actually isn't used by my playbooks (the comments I put in here says this only works on Ansible 2.0 and I'm on Ansible 1.9). But I assume your same comment can apply to the task yml file I do use (fetch-wildfly.yml)

jmazzitelli:
closing - we don't want to merge this anytime soon.

PR #161, opened 2016-03-21
HWKAGENT-63 be more forgiving if hawkular server isn't up yet

jmazzitelli:


ppalaga:
@jmazzitelli When thinking of how to test this, I came across the idea that it would be handy to have `retryWait` configurable as e.g. system props, so that the test do not have to take multiples of 1 minute.

When thinking of it again, why do we actually let the agent crash after `retriesRemaining` retries? The container hosting the Agent will continue to work as usual, but will never (unless rebooted or cli'd) try to reconnect to hawkular (and will not report any single metric), right? Why not let the Agent try to reconnect forever?

The above was for the present PR. What follows is an idea for the future.

Generally, I think that Hawkular server being down (for say up to tens of hours - e.g. Friday evening through Monday morning) should be considered expected. I think that ideally, in case of such Hawkular outages, the Agent should be able to do all its usual work (discover and measure) buffering/storing the data locally until the Hawkular server gets online again. WDYT?

jmazzitelli:
> Why not let the Agent try to reconnect forever?

That is how I originally had it. However, I found out that the agent, while stuck in this initialization state, forbids the Wildfly server from being able to be killed (except with kill -9). Try it - you can't Control-C it, you can't "kill <pid>" it - I think during subsystem initialization, Wildfly doesn't install its shutdown hook yet. So, in other words, if you start a WildFly instance that has an agent in it, and the agent is enabled but the hawkular server is not up, you cannot shutdown that wildfly server unless you do a kill -9. I figure it was best (for now) to have the agent stop within an hour so at least at SOME point, that WildFly instance can be gracefully shutdown (remember, these WildFly instances aren't necessary just an agent, it could be a customer's production instance with the agent added on to it). I plan on looking more into how to fix this, but for now, I just assume not hobble the WildFly instance like this.

jmazzitelli:
> Agent should be able to do all its usual work (discover and measure)
> buffering/storing the data locally until the Hawkular server gets online again. WDYT?

This is exactly how the RHQ/JON agent worked. But we have not had any user requirements or use cases that required us to implement this feature in Hawkular, which is why it hasn't been done yet.  There are difficulties to overcome in order to get such a thing correctly implemented (you have to worry about not buffering/storing so much that you blow up memory or disk space, or consider the amount of time, CPU, etc. it would take to spool days-worth of data from memory/disk to Hawkular Metrics while still collecting current data on time, and doing all this while not adversely affecting the managed resources and the machine they are running on, as examples). I hesitate to implement this until we have definitive requirements for why we would want this and what restrictions we can place on this (e.g. is it OK to buffer just an hour's worth? A day? Does a customer care about several days-old metric data, or just a hour?). In general, if we went days without storing metric data in the server, the usefulness of that old data is reduced - on Monday, users don't care as much about metric data from last Friday compared to knowing the data that is getting collected right now. So we want to be careful in deciding how much data we collect and store.

jmazzitelli:
> retryWait configurable as e.g. system props, so that the test do not have to take multiples of 1 minute.

Agian, something I originally had in the plan but opted not to in order to get this implemented quick. This is something we can add - I was thinking of not using a sysprop but instead using a true attribute in the <subsystem> configuration. In fact I would have two - "retryCount" and "retryWait" so if a person DID want the agent to retry forever, they can just configure retryCount to max_int using the normal way to configure the agent (via standalone.xml) - this would address your first question - why not let the agent retry forever. Having a retryWait can help with tests. We could fallback to a system property in the attribute value if we want: retryWait="${hawkular.agent.retry-wait:60000}".

ppalaga:
@jmazzitelli thanks for the info. I am OK with implementing the proposed improvements later and this PR has green light from me.

jshaughn:
perhaps this  PR comment exchange should be captured in a Jira for agent enhancements?

PR #162, opened 2016-03-23
HWKAGENT-60 use aesh. 

jmazzitelli:
There is one test error here that will pass once that aesh issue is fixed (aesh-348).

do not merge yet


jmazzitelli:
don't merge yet - there is still a bug in aesh that should get fixed first. see jira for more

PR #167, opened 2016-04-22
HWKAGENT-72 be able to select the metrics from prometheus endpoints to store

jmazzitelli:


hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/dynamicprotocol/DynamicEndpointService.java
```diff
@@ -33,16 +36,17 @@
     private final MonitoredEndpoint<DynamicEndpointConfiguration> endpoint;
     private final String feedId;
     private final HawkularWildFlyAgentContext hawkularStorage;
+    private final Collection<Name> metrics; // if not empty, these name the metrics this service should monitor
```
jshaughn, line 39: perhaps this should be pattern based with regex?

jmazzitelli, line 39: yes, I like that. OK, I'll make that change

jshaughn:
Other than the regex comment, LGTM.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/dynamicprotocol/DynamicEndpointService.java
jmazzitelli, line 39: last comment adds the regex feature. i'll merge soon

PR #169, opened 2016-04-26
HWKAGENT-71 Get agent to be installable and runnable inside host controller

jmazzitelli:
this simply allows the agent to be installed in a host controller and run in a host controller.

It runs, but because some DMR resources are missing or in different paths when in host controller vs. standalone, the monitoring fails. We'd just have to figure out what, if anything, we want to monitor in here. But I think that's a separate issue we can work on later. This simply gets the installation correct so it can at least run in the different container environments.

### DIFF
--- a/hawkular-wildfly-agent-installer/src/main/java/org/hawkular/wildfly/agent/installer/TargetConfigInfo.java
+++ b/hawkular-wildfly-agent-installer/src/main/java/org/hawkular/wildfly/agent/installer/TargetConfigInfo.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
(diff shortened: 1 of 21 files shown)


hawkular-wildfly-agent-installer/src/main/java/org/hawkular/wildfly/agent/installer/TargetConfigInfo.java
```diff
@@ -0,0 +1,35 @@
...
+     * @return the XPath that refers to the security-realms element in the config file.
+     */
+    String getSecurityRealmsXPath();
+
+    /**
+     * @return the XPath that refers to the profile element in the config file.
+     */
+    String getProfileXPath();
+
+}
```
jshaughn, line 35: if you were interested in using java8 default methods this might be a decent place to do it since the overriding impls are often the same.

jshaughn:
LOKTM

PR #171, opened 2016-04-28
HWKAGENT-82 full installer module

jmazzitelli:
build a full installer so people can run it without downloading the wf-extension module distro from a running server

### DIFF
--- a/hawkular-wildfly-agent-installer-full/assembly.xml
+++ b/hawkular-wildfly-agent-installer-full/assembly.xml
@@ -1,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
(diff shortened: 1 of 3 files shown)


jmazzitelli:
@jshaughn please peer review. This is simple - it is a new maven module that simply puts the agent wf-extension zip in the root directory of the installer jar - the whole thing is repackaged in a new installer jar that can be executed without requiring a hawkular server to be running in order to be able to install an agent.  You run the artifact this new maven module generates like this, as an example:

java -jar hawkular-wildfly-agent-installer-full-0.17.3.Final-SNAPSHOT.jar --target-location /opt/wildfly --server-url=http://your-server-that-isnt-running:8080 --module-dist classpath:hawkular-wildfly-agent-wf-extension.zip

PR #174, opened 2016-05-04
HWKAGENT-68 change config at runtime

jmazzitelli:
This allows someone to add/remove and enable/disable remote endpoints (remote-dmr/jmx/prometheus) and the local wildfly endpoint (local-dmr).

All other configuration settings will require a restart of the agent subsystem to take effect.

### DIFF
--- a/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/dynamicprotocol/DynamicProtocolService.java
+++ b/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/dynamicprotocol/DynamicProtocolService.java
@@ -22,21 +22,25 @@ import java.util.Map;
 import java.util.Random;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 
+import org.hawkular.agent.monitor.log.AgentLoggers;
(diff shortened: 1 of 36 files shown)


hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/dynamicprotocol/DynamicProtocolService.java
```diff
@@ -51,29 +55,29 @@ public static Builder builder() {
...
-        return endpointServices;
+        synchronized (endpointServices) {
+            return new HashMap<>(endpointServices);
+        }
     }
 
     public void start() {
         ThreadFactory threadFactory = ThreadFactoryGenerator.generateFactory(true,
                 "Hawkular WildFly Agent Dynamic Protocol Service Thread Pool");
         threadPool = Executors.newScheduledThreadPool(1, threadFactory);
```
jshaughn, line 77: Does anything prevent multiple start calls from causing an issue? Or start and stop from stepping on each other?  Maybe it's not a concern and just governed by the callers.

jshaughn:
This is pretty big but it looks OK to me.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/dynamicprotocol/DynamicProtocolService.java
jmazzitelli, line 77: governed by the lifecycle of the agent service - this start is called only when wildfly starts the service. same with stop (when wildfly shuts down the service).

PR #177, opened 2016-05-05
Update the okhttp component from 2.4.0 to 2.7.5

mwringe:
Update to a newer version of the okhttp components.

https://github.com/fabric8io/kubernetes-client requires a newer version to function.

### DIFF
--- a/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/cmd/FeedCommProcessor.java
+++ b/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/cmd/FeedCommProcessor.java
@@ -40,14 +40,16 @@ import org.hawkular.cmdgw.api.Authentication;
 import org.hawkular.cmdgw.api.GenericErrorResponse;
 import org.hawkular.cmdgw.api.GenericErrorResponseBuilder;
 
+import com.squareup.okhttp.MediaType;
+import com.squareup.okhttp.RequestBody;
 import com.squareup.okhttp.Response;
+import com.squareup.okhttp.ResponseBody;
 import com.squareup.okhttp.ws.WebSocket;
 import com.squareup.okhttp.ws.WebSocketCall;
 import com.squareup.okhttp.ws.WebSocketListener;
 
 import okio.Buffer;
 import okio.BufferedSink;
-import okio.BufferedSource;
 
 public class FeedCommProcessor implements WebSocketListener {
     private static final MsgLogger log = AgentLoggers.getLogger(FeedCommProcessor.class);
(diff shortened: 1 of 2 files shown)


hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/cmd/FeedCommProcessor.java
```diff
@@ -185,17 +187,25 @@ public void run() {
...
+                                return WebSocket.BINARY;
+                            }
+
+                            @Override
+                            public void writeTo(BufferedSink bufferedSink) throws IOException {
+                                    emitToSink(messageData, bufferedSink);
+                            }
+                        };
+
+                        FeedCommProcessor.this.webSocket.sendMessage(requestBody);
```
jmazzitelli, line 208: 1. Is there no more need to close anything? In above line 197 (when we had a sink) we needed to close it in finally block. Now I see sink goes away and we have a request body. Just making sure there is nothing we explicitly have to close here. (same for line 252 below - if close is needed here somewhere, it will be needed down there as well)
2. trivial - line 204 is indented improperly - looks like there are 4 extra spaces - checkstyle is probably going to fall on this line?

pom.xml
```diff
@@ -58,7 +58,7 @@
     <version.commons.io>2.4</version.commons.io>
     <version.com.github.dblock>1.5</version.com.github.dblock>
     <version.com.samskivert.jmustache>1.11</version.com.samskivert.jmustache>
-    <version.com.squareup.okhttp>2.4.0</version.com.squareup.okhttp>
+    <version.com.squareup.okhttp>2.7.5</version.com.squareup.okhttp>
```
jmazzitelli, line 61: this is now being managed by the parent pom - see https://github.com/hawkular/hawkular-parent-pom/pull/79
so we should upgrade the parent pom so we can pull that in and not have to manage this version ourselves.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/cmd/FeedCommProcessor.java
mwringe, line 208: For 1) if we created the sink (like before) then we should definitely close it, but since we are not the ones creating the sink we its dangerous to be closing it since we don't know what they want to be doing with the sink afterwards. Eg in the case of multipart messages, they need the sink to remain open since they need to write to it multiple times (https://github.com/square/okhttp/blob/parent-2.7.5/okhttp/src/main/java/com/squareup/okhttp/MultipartBuilder.java#L277) 

For 2) this passes the checkstyle for me either way. I have updated it so that it is more properly formatted.

mwringe: adding message in the hopes that github figures out this conversation is still valid :)

mwringe:
Update to fix the indentation. I added remarks about that we should not be closing the sink ourselves, but github is hiding the comment by default since it thinks that section is outdated :/

jmazzitelli:
don't merge until agent moves up to the latest parent-pom - once that happens, we can use parent-pom to pull in the okhttp version.

ppalaga:
Closing as the commit from here was cherry-picked (and a little bit adapted) in #186

PR #179, opened 2016-05-09
this "deploy" operation should not be here. There is no "Deploy" DMR …

jmazzitelli:
There is no "Deploy" DMR  operation that ExecuteOperationRequest can process.

We already have DeployApplicationRequest that the agent can invoke via the Websocket command interface, but these <operation> metadata is not related to the websocket commands like that.

These DMR operations are only invoked directly over the WildFly management interface via the ExecuteOperationRequest. And there is no "deploy" DMR operation on the root resource in WildFly.


jmazzitelli:
Turns out, the agent does use this - it looks for a "Deploy" operation and if it doesn't see it , assumes that feature is disabled. I think we use this as a security feature - if someone doesn't want to allow the agent to deploy apps to the app server, they can take out that operation and the agent will not deploy anything to that server.

PR #183, opened 2016-05-16
HWKAGENT-90 register the agent feed under all the tenant IDs from all endpoints

jmazzitelli:


hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
```diff
@@ -903,40 +904,63 @@ private void registerFeed(String tenantId) throws Exception {
...
-                    // should we throw an error here or just use the feed ID we were given?
-                    log.debugf("Using feed ID [%s]; make sure the agent doesn't lose its data file", feed.getId());
+            // because we can support multiple tenants, we need to register our feed under all tenants
+            List<AbstractEndpointConfiguration> endpoints = new ArrayList<>();
+            endpoints.addAll(configuration.getDmrConfiguration().getEndpoints().values());
+            endpoints.addAll(configuration.getJmxConfiguration().getEndpoints().values());
+            endpoints.addAll(configuration.getPlatformConfiguration().getEndpoints().values());
+            endpoints.addAll(configuration.getPrometheusConfiguration().getEndpoints().values());
+
+            List<String> tenantIds = new ArrayList<String>();
```
jshaughn, line 914: Set<String>?

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
```diff
@@ -903,40 +904,63 @@ private void registerFeed(String tenantId) throws Exception {
...
+            tenantIds.add(configuration.getStorageAdapter().getTenantId()); // always register agent's global tenant ID
+            for (AbstractEndpointConfiguration endpoint : endpoints) {
+                String tenantId = endpoint.getTenantId();
```
jshaughn, line 917: Does this return a proper default of 'hawkular'?

jshaughn:
Other than the comments, LGTM

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
jmazzitelli, line 917: yes, the setting of the default happens earlier (WildFly API will actually return us the default value if one isn't defined - happens earlier)

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
jmazzitelli, line 914: doesn't matter. List, Set. whateve. Its not used other than here in this local code so it doesn't matter.

jshaughn, line 914: OK, I just figured why make multiple calls if 10 managed servers all used the same tenantId.  if that's not an issue then go ahead and merge.

PR #184, opened 2016-05-16
Add Swarm Distribution for the Agent

jshaughn:
- create hawkular agent fraction
- create swarm jar distribution of the agent
  - provide default config with sysprop overrides
  - provide ability to supply custom config xml (a standalone.xml snippet)

REVIEW ONLY - DO NOT MERGE: SNAPSHOT DEP IN USE !


jmazzitelli:
This requires -Denforcer.skip due to the need to build with a swarm snapshot. We will merge once swarm releases their next release which has a bug fix we need for this PR.

jmazzitelli:
just a reminder that when this gets merged, we should delete the mazz/wf9-swarm branch as that is now obsolete and no longer needed

jshaughn:
@jmazzitelli I think this is ready to merge assuming travis is happy... and then you can delete your old branch as well...

PR #190, opened 2016-05-23
Add default tenantId to the default storage-adapter config

ppalaga:


hawkular-wildfly-agent-feature-pack/src/main/resources/subsystem-templates/hawkular-wildfly-agent.xml
```diff
@@ -30,7 +30,7 @@
                  timeUnits="minutes"/>
 
     <storage-adapter type="HAWKULAR"
-                     tenantId="${hawkular.rest.tenantId}"
+                     tenantId="${hawkular.rest.tenantId:hawkular}"
```
pilhuhn, line 33: I wonder why this has the `.rest` part all over (as with user name and password below). Do we expect othere for the future? @jmazzitelli

ppalaga, line 33: > Do we expect othere for the future?

I do not think there a real expectation of having other `hawkular.*.user` s but I found it good to make it as self explaining as possible what are these variables meant for.

PR #196, opened 2016-05-26
HWKAGENT-98 itests for domain

jmazzitelli:
Even though the test passes, there are still some errors in the domain controller's log that are suspicious. Need to figure out what the issues are.


jmazzitelli:
The issues involve the fact that server-three is not auto-started by default so the agent is just saying it can't collect some details about it, which is to be expected. I think this can get merged.

PR #197, opened 2016-05-31
HAWKULAR-1079 - Moved Agent Download WAR to Agent from Hawkular.

jpkrohling:


jpkrohling:
This includes all the history of the Agent Download WAR from Hawkular plus the necessary changes to fit into the Agent (last 3 commits).

jpkrohling:
@ppalaga last review? Or should I hand it to @jmazzitelli ? :)

hawkular-wildfly-agent-download/src/main/webapp/WEB-INF/web.xml
```diff
@@ -0,0 +1,48 @@
...
+-->
+<web-app xmlns="http://xmlns.jcp.org/xml/ns/javaee"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/web-app_3_1.xsd"
+         version="3.1">
+
+  <security-constraint>
+    <web-resource-collection>
+      <web-resource-name>Installer download endpoint</web-resource-name>
+      <url-pattern>/installer</url-pattern>
```
jmazzitelli, line 28: there are two endpoints for this servlet:

1) /installer -- this downloads the actual installer, with the customized installer config
2) /download -- this downloads the agent's module subsystem extension zip - this endpoint is accessed by the installer

jmazzitelli, line 28: oh wait, but this is security contraints. so we don't want to wrap /download with it. Never mind :)

hawkular-wildfly-agent-download/src/main/webapp/WEB-INF/jboss-deployment-structure.xml
```diff
@@ -0,0 +1,27 @@
...
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<jboss-deployment-structure>
+  <deployment>
+    <dependencies>
+      <module name="org.hawkular.nest" />
+      <module name="org.hawkular.commons.cassandra-driver" />
```
jmazzitelli: do we need cassandra-driver as a dependency? I can't think of why we would need that.

ppalaga:
Ad https://github.com/jpkrohling/hawkular-agent/blob/2b026565efbbda57b5988818ea1047785dca38d2/hawkular-wildfly-agent-download/pom.xml#L68-L69 :
`addDefaultImplementationEntries` are set to `true` in `hawkular-parent` https://github.com/hawkular/hawkular-parent-pom/blob/master/pom.xml#L999-L1014

`addDefaultSpecificationEntries` bring the same kind of info to MANIFEST.MF. Are they somehow useful? If not, the whole `maven-war-plugin` could be removed.

ppalaga:
Is not `@PermitAll` contradictory to what's there in `web.xml` ?

ppalaga:
https://github.com/jpkrohling/hawkular-agent/blob/2b026565efbbda57b5988818ea1047785dca38d2/hawkular-wildfly-agent-download/src/main/java/org/hawkular/component/wildflymonitor/WildFlyAgentServlet.java#L250 I understand this per line properties manipulation as a way to preserve the order of props, their comments and setting null by prepending `#`. Yes, I agree there is no standard API to do this, but maybe using commons-configuration [1] would be less error prone. 

To name the few potentially buggy things I have noticed in WildFlyAgentServlet: 
- property files are always encoded using ISO-8859-1 rather than UTF-8
- values in property files may span over multiple lines using `\`
- `propNameToFind` in https://github.com/jpkrohling/hawkular-agent/blob/2b026565efbbda57b5988818ea1047785dca38d2/hawkular-wildfly-agent-download/src/main/java/org/hawkular/component/wildflymonitor/WildFlyAgentServlet.java#L366 may contain characters with special meaning in regular expressions. `propNameToFind` should therefore be escaped using `...\Q"+ propNameToFind +"\E....`

[1] https://commons.apache.org/proper/commons-configuration/apidocs/org/apache/commons/configuration2/PropertiesConfigurationLayout.html

ppalaga:
No more comments

jpkrohling:
I'll save these comments in a JIRA, as they are not related to the target of this PR, even though I agree those are required changes. The only change I see as related is the `@PermitAll`, which I'm removing.

jpkrohling:
Created https://issues.jboss.org/browse/HAWKULAR-1081 for the two comments from Peter.

hawkular-wildfly-agent-download/src/main/webapp/WEB-INF/jboss-deployment-structure.xml
jpkrohling: ACK, will do.

jpkrohling:
@jmazzitelli , updated PR to remove the `@PermitAll` and dependency on Cassandra. Would you please review/merge?

PR #198, opened 2016-05-31
HWKAGENT-407 mark avail down if can't connect to endpoint

jshaughn:
HWKAGENT-407  when unable to connect to the endpoint, set avails to down
- for remote monitoring this should solve the issue, if you can't connect
  to the endpoint we should assume it's down, and that it's associated
  resources are down.
- for local/embedded dmr this shouldn't happen, because if the server
  goes down so will the agent

### DIFF
--- a/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/api/SamplingService.java
+++ b/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/api/SamplingService.java
@@ -44,6 +44,7 @@ public interface SamplingService<L> {
 
     /**
      * Checks the availabilities defined by {@code instances} and reports them back to the given {@code consumer}.
+     * If the relevant Endpoint is unreachable all instances are reported as {@code Avail.DOWN}.
      *
      * @param instances the availabilities to check
      * @param consumer the consumer to send the results to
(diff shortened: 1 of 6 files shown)


jshaughn:
@jmazzitelli please review.  We may want to wait on any merge for the fix needed to run the itest

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/api/SamplingService.java
```diff
@@ -44,6 +43,8 @@
 
     /**
      * Checks the availabilities defined by {@code instances} and reports them back to the given {@code consumer}.
+     * If the relevant Endpoint is unreachable all instances are reported as {@code Avail.DOWN}.
+
```
jmazzitelli: blank line in the middle of javadoc?

jmazzitelli:
Let's wait for https://issues.jboss.org/browse/HWKMETRICS-407 to be fixed so we can enable the test. If that isn't going to happen in a while, then we can merge this but first write a JIRA to remind us to re-enable this test once that metrics bug is fixed.

jmazzitelli:
@jshaughn so that metrics JIRA isn't really a bug. Can we re-enable that test and then merge this when it goes green?

jshaughn:
@jmazzitelli You're right and the new test is already enabled in this PR.  The entire -Pitests run has passed twice for me locally. I restarted travis again now, I have no idea why it fails but it seems travis-related.  This PR is ready for merge imho.

PR #201, opened 2016-06-03
HWKAGENT-99 rename all attributes so they do not use camel-case

jmazzitelli:
agent subsystem now follows the naming convention of all wildfly subsystems


jshaughn:
That must have been fun to do...

PR #206, opened 2016-06-07
Command cli utility

jmazzitelli:
a simple tool to send json requests to the hawkular server. useful to simulate a ui or other client that wants to send commands to the agent


jshaughn:
Perhaps send out a few examples of how to use it.

PR #208, opened 2016-06-13
HWKAGENT-100 Add agent availability ping

jshaughn:
The agent now sends an availability metric for each of its registered feeds.
The server (in hawkular mode) uses this to determine if the agent is or is
not reporting, and will "backfill" the agent's monitored resources as needed.
- ping-dispatcher-period-secs is a new subsystem attribute setting the
  ping frequency.  It defaults to 60s.
- note, no longer write feedId to a file, instead just re-register feeds to
  ensure all feeds are registered and pinging.

@jmazzitelli Please Review, this is probably close but I could easily have missed something given my new-ness to the code.

### DIFF
--- a/hawkular-swarm-agents/hawkular-swarm-agent-dists/hawkular-swarm-agent-dist/src/main/resources/hawkular-swarm-agent-config.xml
+++ b/hawkular-swarm-agents/hawkular-swarm-agent-dists/hawkular-swarm-agent-dist/src/main/resources/hawkular-swarm-agent-config.xml
@@ -1036,7 +1036,7 @@
 
       <managed-servers>
         <remote-dmr enabled="${hawkular.dmr.enabled:true}"
-                    name="Wildfly"
+                    name="WildFly"
                     host="${hawkular.dmr.host:127.0.0.1}"
                     port="${hawkular.dmr.port:9990}"
                     username="${hawkular.dmr.username:admin}"
--- a/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/scheduler/SchedulerService.java
+++ b/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/scheduler/SchedulerService.java
@@ -18,6 +18,9 @@ package org.hawkular.agent.monitor.scheduler;
 
 import java.util.Collection;
 import java.util.List;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 
 import org.hawkular.agent.monitor.api.InventoryEvent;
 import org.hawkular.agent.monitor.api.InventoryListener;
@@ -33,6 +36,7 @@ import org.hawkular.agent.monitor.storage.AvailBufferedStorageDispatcher;
 import org.hawkular.agent.monitor.storage.AvailDataPoint;
 import org.hawkular.agent.monitor.storage.MetricBufferedStorageDispatcher;
 import org.hawkular.agent.monitor.storage.MetricDataPoint;
+import org.hawkular.agent.monitor.storage.PingStorageDispatcher;
 import org.hawkular.agent.monitor.storage.StorageAdapter;
 
 /**
@@ -45,8 +49,12 @@ public class SchedulerService implements InventoryListener {
     private final Diagnostics diagnostics;
     private final MeasurementScheduler<Object, MetricType<Object>, MetricDataPoint> metricScheduler;
     private final MeasurementScheduler<Object, AvailType<Object>, AvailDataPoint> availScheduler;
+    private final ScheduledThreadPoolExecutor pingScheduler;
     private final MetricBufferedStorageDispatcher metricStorage;
     private final AvailBufferedStorageDispatcher availStorage;
+    private final PingStorageDispatcher pingStorage;
+
+    private ScheduledFuture<?> pingJob;
 
     protected volatile ServiceStatus status = ServiceStatus.INITIAL;
 
@@ -58,7 +66,7 @@ public class SchedulerService implements InventoryListener {
         // metrics for our own internals
         this.diagnostics = diagnostics;
 
-        // create the schedulers - we use two: one for metric collections and one for avail checks
+        // create the schedulers - we use three: one for metric collections, one for avail checks and one for feed pings
         this.metricStorage = new MetricBufferedStorageDispatcher(configuration, storageAdapter, diagnostics);
         this.metricScheduler = MeasurementScheduler.forMetrics("Hawkular-WildFly-Agent-Scheduler-Metrics",
                 metricStorage);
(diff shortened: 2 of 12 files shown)


jmazzitelli:
I'll review tomorrow, but I already have a comment :)

"ping-dispatcher-period-secs" - I don't know (and I'm sure customers won't know) what "dispatcher" means. I would think naming this something more simpler like "ping-period-secs" would be good without the confusing "dispatcher" term. Or maybe "heartbeat-ping-period-secs" or something more descriptive than the nebulous "dispatcher".

jshaughn:
OK, I'll change it, I guess I was just caught up in parity with the 
existing dispatchers...

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/scheduler/SchedulerService.java
```diff
@@ -58,14 +66,17 @@ public SchedulerService(
...
         this.metricStorage = new MetricBufferedStorageDispatcher(configuration, storageAdapter, diagnostics);
         this.metricScheduler = MeasurementScheduler.forMetrics("Hawkular-WildFly-Agent-Scheduler-Metrics",
                 metricStorage);
 
         this.availStorage = new AvailBufferedStorageDispatcher(configuration, storageAdapter, diagnostics);
         this.availScheduler = MeasurementScheduler.forAvails("Hawkular-WildFly-Agent-Scheduler-Avail",
                 availStorage);
+
+        this.pingStorage = new PingStorageDispatcher(configuration, storageAdapter, diagnostics);
+        this.pingScheduler = new ScheduledThreadPoolExecutor(1);
```
jmazzitelli: pass in a thread factory here so we can give the thread a name (the default will be a generic "ThreadPool-01" or something like that, and its hard to see it in thread dumps.) and make it a daemon thread so it won't hang up the VM exiting if it gets stuck. So something like:

```
    ThreadFactory threadFactory = ThreadFactoryGenerator.generateFactory(true, "Hawkular-WildFly-Scheduler-Ping");
```

And pass that thread factory to the ScheduledThreadPoolExecutor constructor

jmazzitelli:
Two things -

1) if this is in METRICS-ONLY mode, do we still send these pings? I didn't see if we do anything to disable pings in in metrics-only mode.

2) it is possible for new managed servers to be added at runtime. Right now, if someone adds a new managed server (say, remote-dmr) with a separate tenant ID, that tenant ID's ping will not go out unless the agent is restarted. If we do not plan on fixing this, we need to write a JIRA to remind ourselves to fix this later. Also, you might want to see if you can refactor some of this to make it easier to incorporate that fix when we do it. Looking at the code, I don't know how easy it will be to add a new tenant ID to the set of tenant IDs that the ping dispatcher uses. Take a look at:

org.hawkular.agent.monitor.extension.RemoteDMRAdd.performRuntime

for the code that is executed when a new remote-dmr is added at runtime. Then look at:

org.hawkular.agent.monitor.extension.RemoteDMRDefinition.registerAttributes

for the code that is executed when an existing remote-dmr is enabled or disabled.

We would need to add code in those places to add or remove tenant ID from the set of tenant IDs the dispatcher uses. In addition, there are other managed servers that allow for this too (remote-jmx and remote-prometheus - I suspect we'll need to do the same thing there - those two managed servers have almost identical code in their "Add" and "Definition" classes.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
```diff
@@ -864,6 +866,9 @@ private void startScheduler() throws Exception {
             schedulerConfig.setMetricDispatcherMaxBatchSize(this.configuration.getMetricDispatcherMaxBatchSize());
             schedulerConfig.setAvailDispatcherBufferSize(this.configuration.getAvailDispatcherBufferSize());
             schedulerConfig.setAvailDispatcherMaxBatchSize(this.configuration.getAvailDispatcherMaxBatchSize());
+            schedulerConfig.setPingDispatcherPeriodSeconds(this.configuration.getPingDispatcherPeriodSeconds());
+            schedulerConfig.setPingDispatcherFeedId(this.feedId);
```
jmazzitelli: why don't we just make this "setFeedId" since its possible in the future other schedulers might need to know the agent's feed ID. Change the getter too to be "getFeedId"

jshaughn:
> 1) if this is in METRICS-ONLY mode, do we still send these pings? I didn't see if we do anything to disable pings in in metrics-only mode.

We still ping by default. We don't have to, I wasn't sure, really. I didn't know if it was useful to just provide agent avail for its own worth.  It would be shut off in the same way it would be for hawkular mode, set the ping period to 0.

I'll look and see if I can easily set it to 0 for metrics_only mode, if not I'll leave it as is for now.

jshaughn:
Actually, there was a bug such that we were not pinging in Metrics_Only mode.  After some discussion, we'll fix that bug and make things consistent.  By default, regardless of mode well push ping avails at 60s.  Users can easily turn it off for either mode.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
```diff
@@ -976,6 +958,31 @@ public String getTenantId() {
         return this.storageAdapter.getStorageAdapterConfiguration().getTenantId();
     }
 
+    /**
+     * @return tenant ID of the agent and its monitored endpoints - if the agent has not started, empty set is returned
+     */
```
jmazzitelli: that's incorrect javadoc

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
```diff
@@ -976,6 +958,31 @@ public String getTenantId() {
         return this.storageAdapter.getStorageAdapterConfiguration().getTenantId();
     }
```
jmazzitelli: replace the content of this method with this:
        return this.configuration.getStorageAdapter().getTenantId()

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
```diff
@@ -984,6 +958,31 @@ public String getTenantId() {
...
 
+    /**
+     * @return tenant ID of the agent and its monitored endpoints - if the agent has not started, empty set is returned
+     */
+    public Set<String> getTenantIds() {
+        Set<String> tenantIds = new HashSet<String>();
+        if (this.storageAdapter == null) {
```
jmazzitelli: remove this if-stmt

hawkular-swarm-agents/hawkular-swarm-agent-dists/hawkular-swarm-agent-dist/src/main/resources/hawkular-swarm-agent-config.xml
```diff
@@ -1036,7 +1036,7 @@
 
       <managed-servers>
         <remote-dmr enabled="${hawkular.dmr.enabled:true}"
-                    name="Wildfly"
+                    name="WildFly"
```
jmazzitelli, line 1039: this doesn't have anything to do with this PR.... ah, what the heck, its Tuesday.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
jmazzitelli: ignore that ... I didn't read the javadoc fully

PR #212, opened 2016-06-16
do not include resoure types as root types if the configuration is bad

jmazzitelli:
You will see errors logs when this happens now, and the type won't be considered a root type:

13:53:24,501 ERROR [org.hawkular.agent.monitor.inventory.ResourceTypeManager](Hawkular WildFly Agent Full Discovery Scan-1) HAWKMONITOR010074: The resource type [Datasource] is missing a parent. Make sure at least one of these resource types are defined and enabled: [WildFly Server, Domain WildFly Server]


jmazzitelli:
this is for : https://issues.jboss.org/browse/HWKAGENT-113

PR #215, opened 2016-06-22
put back the Hostname property name to what it was before

jmazzitelli:


theute:
The property name is different if it's in standalone or domain mode, is that expected ?

(Domain mode still uses "Qualified Host Name")

PR #221, opened 2016-06-24
use the new wildfly maven plugin core API for deploy/undeploy

jmazzitelli:
Need to wait for a couple things to be fixed in the wildfly plugin core API before we can merge.

Once version 1.1.0.Alpha10 is released, we'll need to upgrade parent-pom to reference it, then we need to bump up the agent's parent to use that parent pom version, then we should be able to merge.

Right now, itests for this branch will fail due to deployment enablement not working due to bug in maven plugin core api.


jmazzitelli:
Once parent pom version 45 is released (which should define maven plugin version property 1.1.0.Alpha10) this should be able to go green and then we can merge.

In other words, wait for this to be merged and parent-pom 45 to be released with it: https://github.com/hawkular/hawkular-parent-pom/pull/83

jmazzitelli:
This is building locally on my box, however, for some reason travis cannot download the new 1.1.0.Alpha10 release of the maven plugin core, even after I purged travis' cache. I don't know what the problem is, but it is a travis issue.

jshaughn:
LGTM

PR #222, opened 2016-06-27
HWKAGENT-116 - retry when registering feed

jmazzitelli:


hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
```diff
@@ -919,47 +922,69 @@ private void registerFeed(Set<String> tenantIds) throws Exception {
...
-                            log.infoUsingFeedId(feed.getId(), tenantId);
+                boolean keepRetrying = true; // assume it will be a failure that we can retry on
+                do {
+                    try {
+                        // note that we always retry if newCall.execute throws an exception
+                        Response httpResponse = httpclient.newCall(request).execute();
+
+                        try {
+                            // HTTP status of 201 means success; 409 means it already exists, anything else is an error
+                            if (httpResponse.code() == 201) {
```
jshaughn, line 949: We should use a switch statement here. Not a PR blocker.

PR #225, opened 2016-06-29
Add 'Name' and 'Node Name' resource configs

jshaughn:
resource config.


jmazzitelli:
I think we need these also on the "Domain WildFly Server" resource type.

jshaughn:
+1, added to domain wildfly server

jshaughn:
Whoops, reverted the duplicate attr additions.

PR #227, opened 2016-07-08
HWKAGENT-79  metric tags

jmazzitelli:
right now this only supports defining tags per managed server. So for each remote-dmr, remote-jmx, etc, you can define a set of tags for each metric collected under that managed server. You can add tokens to the labels (like %MetricTypeUnits or %ResourceName).

This does not allow you to define tags for specific metric types. I'm thinking about how it would be possible to support that.


jmazzitelli:
note that this is based on the branch from the PR https://github.com/hawkular/hawkular-agent/pull/226 

once that is merged into master, we can rebase this PR branch on master

PR #230, opened 2016-07-13
hwkagent-1097 - add home directory config prop

jmazzitelli:


jmazzitelli:
another bogus travis failure. passes on my box. I need this merged, and since its a fairly trivial change I'm merging now.

PR #231, opened 2016-07-16
Differentiate int and float, add a 'required' attribute.

pilhuhn:
Followup to my previous pull to differentiate int and float and to add a 'required' attribute for a parameter.
Also populate a few more operations.

### DIFF
--- a/hawkular-wildfly-agent/src/main/resources/schema/hawkular-agent-monitor-subsystem.xsd
+++ b/hawkular-wildfly-agent/src/main/resources/schema/hawkular-agent-monitor-subsystem.xsd
@@ -399,6 +399,7 @@
     <xs:attribute name="name"          type="xs:string" use="required"/>
     <xs:attribute name="type"          type="operationParamTypeType" default="string"/>
     <xs:attribute name="default-value" type="xs:string" />
+    <xs:attribute name="required"      type="xs:boolean" default="false" use="optional"/>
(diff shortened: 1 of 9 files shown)


hawkular-wildfly-agent/src/main/resources/schema/hawkular-agent-monitor-subsystem.xsd
```diff
@@ -394,14 +394,16 @@
     <xs:attribute name="name"          type="xs:string" use="required"/>
     <xs:attribute name="type"          type="operationParamTypeType" default="string"/>
     <xs:attribute name="default-value" type="xs:string" />
+    <xs:attribute name="required"      type="xs:boolean" default="false" use="optional"/>
```
jmazzitelli: to make the xsd less noisy, we do not specify "use="optional"" on any xs:attribute since that is the default. Remove "use" attribute here since optional is inferred.

jmazzitelli:
There are now two .xml agent config files we have to keep in sync. You already changed hawkular-wildfly-agent.xml, but don't forget to change the swarm config file too: 

https://github.com/hawkular/hawkular-agent/blob/master/hawkular-swarm-agents/hawkular-swarm-agent-dists/hawkular-swarm-agent-dist/src/main/resources/hawkular-swarm-agent-config.xml

jmazzitelli:
other than my two comments this looks good. so once those small changes are made, this can be merged

PR #232, opened 2016-07-19
Allow to pass in the agent name via -Dhawkular.agent.name

pilhuhn:
Default value is still 'Local'


jshaughn:
Unrelated travis failure. LGTM.

jmazzitelli:
this needs to be reverted. wildfly does NOT allow expressions in "name" attributes. I tried this months ago and found it out the hard way :)

PR #235, opened 2016-07-25
HWKAGENT-121 mark avail as down if we fail to get the value

jmazzitelli:


jmazzitelli:
this passes on my box, travis is just annoying. I need someone to look this over and merge. Its a simple change - just wrapped some code around a try-catch block

jshaughn:
LGTM

jshaughn:
Travis passed but then hung and the status was not updated here.  Merging...

PR #237, opened 2016-07-29
Related to Hawkular-1101

jshaughn:
- By default, don't perform "set-avail-on-shutdown" actions for anything
- Update descriptions for "set-avail-on-shutdown" to state that it should be used only in METRIC mode.
- For a remotely monitored resource, set the following property on
  its avail metric in inventory: hawkular-services.monitoring-type:remote.
  This tells the backfiller to set the avail to UNKNOWN for remotely managed
  resource avails, the default (or if the prop is set to "local") is DOWN.


jmazzitelli:
LTGM, though don't know why travis failed. don't merge until we are sure this isn't a real error (I hate travis)

jmazzitelli:
bump up inventory to its .3 version in this PR

PR #243, opened 2016-08-18
HWKAGENT-118 use formal op parameters

jmazzitelli:
this tests that the agent is storing operation parameters using the "formal" inventory way of parameterTypes data entities. Currently, the agent also stores these in general properties - we'll remove that in the future when clients are able to convert to using the new way (e.g. the ruby client)


jshaughn:
LGTM, unrelated travis hangs but builds locally

PR #244, opened 2016-08-18
HWKAGENT-130 remove op parameters from general properties

jmazzitelli:


jmazzitelli:
do not merge this yet. we'll want to give some time to allow clients and tools to update themselves so they do not rely on the general properties that this PR removes.

jshaughn:
LGTM

jmazzitelli:
This change has been announced on the dev mailing list here: http://lists.jboss.org/pipermail/hawkular-dev/2016-August/003183.html

jshaughn:
I think only the gem needs to be updated.  This can be merged and the Gem will follow in short order.

PR #245, opened 2016-08-19
HWKAGENT-132 remove resources after discovery

jmazzitelli:
this looks to be working. I need some itests for it though.

https://issues.jboss.org/browse/HWKAGENT-132

### DIFF
--- a/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/inventory/ResourceManager.java
+++ b/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/inventory/ResourceManager.java
@@ -28,6 +28,7 @@ import java.util.Set;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.stream.Collectors;
 
 import org.hawkular.agent.monitor.log.AgentLoggers;
 import org.hawkular.agent.monitor.log.MsgLogger;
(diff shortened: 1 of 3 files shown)


hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/inventory/ResourceManager.java
```diff
@@ -415,6 +416,29 @@ public int size(Resource<L> relativeTo) {
     }
 
     /**
+     * @param filter if not null, any resources whose IDs are in here will not exist in the returned collection
+     * @return a list of the resources in this resource manager masked by the given IDs
+     *         (that is, if an ID in this resource manager matches an ID found in the filter collection,
+     *         the resource whose ID it is will not be found in the returned collection).
+     *         DO NOT THINK THIS RETURNS ONLY THOSE RESOURCES WHOSE IDS MATCH THOSE IN FILTER, IT IS THE OPPOSITE.
+     */
+    public List<Resource<L>> getAllResources(Collection<ID> filter) {
```
jshaughn, line 425: A better method or param name would preclude the need for this yelling comment, but I'll let it slide :)

jmazzitelli, line 425: I must have named and renamed that method and param 3 or 4 times - I couldn't figure out a good one that still wouldn't cause the same confusion without the method name being as long as that comment :)

jshaughn:
OK, this looks good and seems relatively efficient as it mainly deals with ids to manipulate the existing inventory, iiuc.

PR #246, opened 2016-08-22
HWKAGENT-133 platform resource types and their metrics should have more unique IDs

jmazzitelli:


jshaughn:
This is probably OK but before releasing (or maybe merging) it may be worthwhile to post changes like this on dev-list, just to make sure there is no downstream downstream impact in the provider code.

jmazzitelli:
already in the process of writing the email

jmazzitelli:
> > post changes like this on dev-list
> 
> already in the process of writing the email

http://lists.jboss.org/pipermail/hawkular-dev/2016-August/003176.html

jshaughn:
A quick github search didn't reveal any hits for the old names, except in miq for some other providers... LGTM.

PR #249, opened 2016-08-24
HWKAGENT-131 better resource types sync

jmazzitelli:


jmazzitelli:
do not merge yet. some weird things are cropping up with these changes around h-inventory - such as getting an inventory exception "java.lang.IllegalStateException: Entity with path: /t;itest-rest-tenant/f;ef20d41b-fa4e-4b98-a7e6-9171aff4823b was associated with more than 1 hash node. That is a bug."

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/storage/AsyncInventoryStorage.java
```diff
@@ -527,30 +544,42 @@ public void shutdown() {
...
 
-        // Note that it is possible for a endpoint to define multiple root resources.
-        // We have to sync each root resource separately.
-        blueprints.forEach((r, bp) -> performResourceSync(bp, tenantIdToUse, event.getResourceManager().size(r)));
-    }
+        // Since we know types never change during the lifetime of the agent, we don't have to process
+        // types that have already been flagged as having been persisted.
+        // We first must persist metric types then resource types (remember, there are no hierarchies,
+        // all metric types and resource types are peers to one another).
+        List<ResourceType<L>> allResourceTypes = resourceTypeManager.getResourceTypesBreadthFirst()
```
jshaughn, line 561: Perhaps 'unpersistedResourceTypes' would have been a more clear name, but whatevs.  Look at mazz getting all stream-y here...

jshaughn:
This incremental approach is both cleaner and better.  Will merge if Travis passes and DNM is removed.

jmazzitelli:
After my latest 1-line commit (synchronize a method), the full test suite passed twice in a row on my box. I think this might be a valid (albeit temporary) workaround to the concurrency issues happening in the server. If the tests pass here, we could merge this to master.

jmazzitelli:
wow, it worked :) I'm going to remove Do No Merge label. I think this should avoid the concurrency issues that we saw in inventory.

PR #250, opened 2016-09-06
provide a yaml config file

jmazzitelli:
this only supports a "read only" yaml configuration file. it must exist in the current working directory for swarm. For full agent, it must exist in the WildFly configuration/ directory.

This is just a PoC. Do not merge yet.


jmazzitelli:
We are not going to do this.

PR #254, opened 2016-09-19
hwkagent-139 support string metrics

jmazzitelli:


hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/protocol/EndpointService.java
```diff
@@ -443,21 +445,45 @@ public void measureMetrics(Collection<MeasurementInstance<L, MetricType<L>>> ins
...
-                        double num = toDouble(item);
-                        value += num;
+                Object metricValue; // will be either a String or Double
+                if (instance.getType().getMetricType() == org.hawkular.metrics.client.common.MetricType.STRING) {
+                    StringBuilder svalue = new StringBuilder();
+                    if (o instanceof List<?>) {
+                        /* aggregate */
+                        List<?> list = (List<?>) o;
+                        for (Object item : list) {
+                            svalue.append(String.valueOf(item));
```
jshaughn, line 458: This is sort of weird, I guess we're just trying to do something reasonable with a list of Strings.  Perhaps CSV would be a better option than just stringing it all together with no delims.

Also, would be a little more efficient to just use the StringBuilder for the List case.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/storage/NumericMetricDataPoint.java
```diff
@@ -0,0 +1,51 @@
...
+     * If the underlying data point was null, a 0.0 is returned.
+     *
+     * @return non-null double value for the metric data point.
+     */
+    public Double getMetricValue() {
+        Object v = super.getMetricValue();
+        if (v instanceof Number) {
+            return ((Number) v).doubleValue();
+        } else {
+            return (v != null) ? (Double.valueOf(v.toString())).doubleValue() : Double.valueOf(0.0);
```
jshaughn: Any chance of NumberFormatException?  Also, I think you can drop the .doubleValue() after Double.valueOf(v.toString())

jmazzitelli:
1. Aggregating string metrics creates a CSV string.
2. dropped doubleValue call
3. With the current implementation, I don't think it's possible for a NumberFormatException to occur (the Double is coming from an actual double primitive in the current code) but nevertheless, even if it is possible, I want to leave it like this. I _want_ a runtime NumberFormatException to be thrown here if the string value cannot be parsed. So it's doing the correct thing.

PR #256, opened 2016-09-23
Inv 0.19.2.final

metlos:
Another attempt at getting this past Travis..


jshaughn:
@metlos, @jmazzitelli,  travis gave up after 2 hours and 243 retries, so it doesn't seem like waiting longer is the answer to whatever problem there is...  I can restart the job but not sure if that is useful...

pilhuhn:
On 23 Sep 2016, at 14:16, Jay Shaughnessy wrote:

> @metlos, @jmazzitelli,  travis gave up after 2 hours and 243 retries, 
> so it doesn't seem like waiting longer is the answer to whatever 
> problem there is...  I can restart the job but not sure if that is 
> useful...

Not at all.

I talked to Lukas earlier
- apparently the agent (or its counterpart) is trying to inform the bus 
  that it is shutting down, which fails as the topic is gone
- It may be that the TCP linger option is on, which will keep port 8080 
  in close wait, while something else is trying to start up on the same 
  port

Perhaps we can better order subsystem startup/shutdown so that things 
are torn down in the order we want.
It may or may not help to wait a few seconds between test runs that 
start/stop WildFly. We could also try to use different port-offsets so 
that a new server does not step on a still open 8080

pilhuhn:
What is the status of this one @metlos ?

metlos:
I stopped looking into this actually...

pilhuhn:
So it can be closed?

metlos:
It depends on @jmazzitelli if he wants to use Travis for his CI... ;)

jmazzitelli:
we got it to go green on inventory 1.0.0.Final in another PR. closing this

PR #270, opened 2016-11-24
Check if metrics is really up or still starting.

pilhuhn:
If metrics returns 200 OK, we still need to check if it is still starting.

HWKAGENT-150

### DIFF
--- a/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
+++ b/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
@@ -21,6 +21,7 @@ import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
 import java.io.File;
+import java.io.IOException;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
@@ -106,6 +107,7 @@ import org.jboss.msc.value.InjectedValue;
 
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.ScheduledReporter;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.squareup.okhttp.OkHttpClient;
 import com.squareup.okhttp.Request;
 import com.squareup.okhttp.Response;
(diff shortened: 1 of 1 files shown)


hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
```diff
@@ -958,6 +965,26 @@ private void waitForHawkularServer() throws Exception {
     }
 
     /**
+     * If the server returns a 200 OK, we still need to check the content if the server
+     * is really up. This is explained here: https://twitter.com/heiglandreas/status/801137903149654017
+     * @param body_string String representation of the body
+     * @return true if it is really up, false otherwise (still starting).
+     */
+    private boolean check_really_up(String body_string) {
```
jmazzitelli: You've been programming in Ruby too long. :) Standards for method names are camel case - please rename to checkReallyUp() and "body_string" to "bodyString". Same on line 914 (body_string == bodyString)

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
```diff
@@ -909,8 +911,13 @@ private void waitForHawkularServer() throws Exception {
...
                     log.debugf("Hawkular Metrics is not ready yet: %d/%s", response.code(), response.message());
                 } else {
-                    log.debugf("Hawkular Metrics is ready: %s", response.body().string());
-                    break;
+                    String bodyString = response.body().string();
+                    if (checkReallyUp(bodyString)) {
+                        log.debugf("Hawkular Metrics is ready %s", bodyString);
+                        break;
+                    } else {
+                        log.debugf("Hawkular Metrics is still starting", bodyString);
```
jmazzitelli: Sorry @pilhuhn  - I just noticed this when I was about to press the merge button. Notice this debugf call passes in "bodyString" but you don't have a %s in the log message, so its not used. Need to fix that. Other than that, this is ready to be merged once that is fixed. So change this line to:

````
log.debugf("Hawkular Metrics is still starting: %s", bodyString);
````

While you are here, probably also good to add a colon to the previous debug log message to separate the static message with the value of bodyString - line 916:

````
log.debugf("Hawkular Metrics is ready: %s", bodyString);
````

PR #271, opened 2016-11-25
[WIP] Migrating agent to use jboss-as compatible API

josejulio:
This is work in progress.

- [x] Have it running on an EAP6 standalone.
- [x] Test it on EAP7
- [x] Find replacement for `PathAddress.parseCLIStyleAddress()`
   - Had to create a new class `PathAddressExtension` with a copy of the wildfly methods with minor modification (see comments in code)
- [x] Find replacement for `OperationContext.getCurrentAddressValue()`
- [x] ~~Fix tests POMs that are complaining about not able to resolve:~~
- [x] Migrate other org.wildfly dependencies (from hawkular-dmr-client)
- [x] Use reflection to call `subsystem.setHostCapable();`
- [ ] Test it and see what else might fail (and update this list).
- [x] ~~Downgrade dependencies to match EAP6~~ 
  - Downgraded but had some issues, it seems easier to build against wildfly.
- [x] Check about the NosuchMethodException  https://github.com/hawkular/hawkular-agent/pull/271#issuecomment-264224063
- [x] Use EAP7 API if running on an EAP7-based instance.
- [x] enable JNDI stuff
- [ ] Look into supporting expressions.
  - To do so, I would need to first fetch the expresion, and then request a :resolve-expression operation on the server/domain, depending the context.
- [ ] Squash / tidy up the commits

### DIFF
--- a/hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
+++ b/hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
+ * Copyright 2015-2017 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -17,6 +17,7 @@
 package org.hawkular.dmr.api;
 
 import java.io.IOException;
+import java.lang.reflect.Field;
 import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -529,7 +530,20 @@ public class OperationBuilder implements SubsystemDatasourceConstants, Subsystem
 
         @SuppressWarnings("unchecked")
         public T resolveExpressions(boolean resolveExpressions) {
-            baseNode.get(ModelDescriptionConstants.RESOLVE_EXPRESSIONS).set(resolveExpressions);
+            try {
+                Field resolveExpressionField = ModelDescriptionConstants.class.getField("RESOLVE_EXPRESSIONS");
+                String resolveExpressionFieldValue = (String) resolveExpressionField.get(null);
+                baseNode.get(resolveExpressionFieldValue).set(resolveExpressions);
+            } catch (NoSuchFieldException e) {
+                // Attempt to set the resolve-expressions field
+                baseNode.get("resolve-expressions").set(resolveExpressions);
+                if (resolveExpressions) {
+                    log.warnf("The local app server instance does not support resolving expressions.");
+                }
+            } catch (IllegalAccessException e) {
+                throw new RuntimeException(e);
+            }
+
             return (T) this;
         }
 
--- a/hawkular-wildfly-agent-feature-pack/src/main/resources/modules/system/add-ons/hawkular-agent/org/hawkular/agent/main/module.xml
+++ b/hawkular-wildfly-agent-feature-pack/src/main/resources/modules/system/add-ons/hawkular-agent/org/hawkular/agent/main/module.xml
@@ -1,6 +1,6 @@
 <!--
 
-    Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
+    Copyright 2015-2017 Red Hat, Inc. and/or its affiliates
     and other contributors as indicated by the @author tags.
 
     Licensed under the Apache License, Version 2.0 (the "License");
@@ -72,6 +72,12 @@
     <artifact name="${httpmime}" />
      -->
 
+    <!-- EAP6 -->
+    <artifact name="${com.fasterxml.jackson.core:jackson-core}"/>
+    <artifact name="${com.fasterxml.jackson.core:jackson-databind}"/>
+    <artifact name="${com.fasterxml.jackson.core:jackson-annotations}"/>
+    <artifact name="${org.glassfish:javax.json}"/>
+
   </resources>
 
   <dependencies>
@@ -96,11 +102,12 @@
     <module name="org.jboss.logmanager" services="import"/>
     <module name="org.jboss.threads"/>
     <module name="org.slf4j"/>
-    <module name="org.wildfly.security.manager"/>
+    <!-- <module name="org.wildfly.security.manager"/> --> <!-- Not included with EAP6 -->
     <module name="sun.jdk"/> <!-- codahale needs sun.misc.Unsafe -->
-    <module name="com.fasterxml.jackson.core.jackson-core"/>
-    <module name="com.fasterxml.jackson.core.jackson-databind"/>
-    <module name="com.fasterxml.jackson.core.jackson-annotations"/>
+    <!-- Not included with EAP6 -->
+    <!-- <module name="com.fasterxml.jackson.core.jackson-core"/> -->
+    <!-- <module name="com.fasterxml.jackson.core.jackson-databind"/> -->
+    <!-- <module name="com.fasterxml.jackson.core.jackson-annotations"/> -->
 
     <system export="true">
       <paths>
--- a/hawkular-wildfly-agent/pom.xml
+++ b/hawkular-wildfly-agent/pom.xml
@@ -35,17 +35,20 @@
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-core</artifactId>
-      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-databind</artifactId>
-      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-annotations</artifactId>
-      <scope>provided</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.glassfish</groupId>
+      <artifactId>javax.json</artifactId>
+      <version>1.0.4</version>
     </dependency>
 
     <dependency>
@@ -113,6 +116,12 @@
     <dependency>
       <groupId>org.jboss.logging</groupId>
       <artifactId>jboss-logging</artifactId>
+      <!-- We need to hardcode this version to ensure that calls like LOG.debugf("x", 1) don't link to
+      void debugf(String format, int arg);
+      Thus breaking compatibility with EAP6 which doesn't have said method.
+      Instead we use 3.1.4.GA to ensure it uses void debugf(String format, Object param1); which is provided
+      on both EAP6 and EAP7 see: https://github.com/hawkular/hawkular-agent/pull/271#discussion_r97190787 -->
+      <version>3.1.4.GA</version>
       <scope>provided</scope>
     </dependency>
 
--- a/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/RemoteDMRAdd.java
+++ b/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/RemoteDMRAdd.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
+ * Copyright 2015-2017 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -26,6 +26,7 @@ import org.hawkular.agent.monitor.protocol.dmr.DMRNodeLocation;
 import org.hawkular.agent.monitor.protocol.dmr.DMRSession;
 import org.hawkular.agent.monitor.service.MonitorService;
 import org.hawkular.agent.monitor.util.Util;
+import org.hawkular.agent.monitor.util.WildflyCompatibilityUtils;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
 import org.jboss.dmr.ModelNode;
@@ -42,7 +43,6 @@ public class RemoteDMRAdd extends MonitorServiceAddStepHandler {
     @Override
     protected void performRuntime(OperationContext context, ModelNode operation, ModelNode model)
             throws OperationFailedException {
-
         if (context.isBooting()) {
             return;
         }
@@ -53,7 +53,7 @@ public class RemoteDMRAdd extends MonitorServiceAddStepHandler {
         }
 
         MonitorServiceConfiguration config = Util.getMonitorServiceConfiguration(context);
-        String newEndpointName = context.getCurrentAddressValue();
+        String newEndpointName = WildflyCompatibilityUtils.getCurrentAddressValue(context, operation);
 
         // Register the feed under the tenant of the new managed server.
         // If endpoint has a null tenant then there is nothing to do since it will just reuse the agent's tenant ID
--- a/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/SubsystemExtension.java
+++ b/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/SubsystemExtension.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
+ * Copyright 2015-2017 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -16,6 +16,7 @@
  */
 package org.hawkular.agent.monitor.extension;
 
+import org.hawkular.agent.monitor.util.WildflyCompatibilityUtils;
 import org.jboss.as.controller.Extension;
 import org.jboss.as.controller.ExtensionContext;
 import org.jboss.as.controller.SubsystemRegistration;
@@ -58,7 +59,7 @@ public class SubsystemExtension implements Extension {
                 MAJOR_VERSION, MINOR_VERSION, MICRO_VERSION);
 
         // This subsystem should be runnable on a host
-        subsystem.setHostCapable();
+        WildflyCompatibilityUtils.subsystemSetHostCapable(subsystem);
 
         final ManagementResourceRegistration registration = subsystem
                 .registerSubsystemModel(SubsystemDefinition.INSTANCE);
--- a/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
+++ b/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
+ * Copyright 2015-2017 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -73,6 +73,7 @@ import org.hawkular.agent.monitor.storage.InventoryStorageProxy;
 import org.hawkular.agent.monitor.storage.MetricStorageProxy;
 import org.hawkular.agent.monitor.storage.StorageAdapter;
 import org.hawkular.agent.monitor.util.Util;
+import org.hawkular.agent.monitor.util.WildflyCompatibilityUtils;
 import org.hawkular.inventory.api.model.Feed;
 import org.jboss.as.controller.ControlledProcessState;
 import org.jboss.as.controller.ControlledProcessStateService;
@@ -83,7 +84,6 @@ import org.jboss.as.domain.management.SecurityRealm;
 import org.jboss.as.domain.management.security.SSLContextService;
 import org.jboss.as.host.controller.DomainModelControllerService;
 import org.jboss.as.host.controller.HostControllerEnvironment;
-import org.jboss.as.naming.ImmediateManagedReferenceFactory;
 import org.jboss.as.naming.ManagedReferenceFactory;
 import org.jboss.as.naming.ServiceBasedNamingStore;
 import org.jboss.as.naming.deployment.ContextNames;
@@ -94,6 +94,7 @@ import org.jboss.as.server.ServerEnvironment;
 import org.jboss.as.server.ServerEnvironmentService;
 import org.jboss.as.server.Services;
 import org.jboss.logging.Logger;
+import org.jboss.msc.inject.Injector;
 import org.jboss.msc.service.AbstractServiceListener;
 import org.jboss.msc.service.Service;
 import org.jboss.msc.service.ServiceBuilder;
@@ -177,7 +178,8 @@ public class MonitorService implements Service<MonitorService> {
                         port = socketBinding.getAbsolutePort();
                     } else {
                         OutboundSocketBinding serverBinding = serverOutboundSocketBindingValue.getValue();
-                        address = serverBinding.getResolvedDestinationAddress().getHostName();
+                        address = WildflyCompatibilityUtils.
+                                outboundSocketBindingGetResolvedDestinationAddress(serverBinding).getHostName();
                         port = serverBinding.getDestinationPort();
                     }
                     String protocol = (bootStorageAdapter.isUseSSL()) ? "https" : "http";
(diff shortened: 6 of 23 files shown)


hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/SubsystemExtension.java
```diff
@@ -58,7 +58,7 @@ public void initialize(ExtensionContext context) {
                 MAJOR_VERSION, MINOR_VERSION, MICRO_VERSION);
 
         // This subsystem should be runnable on a host
-        subsystem.setHostCapable();
+        //subsystem.setHostCapable();
```
jmazzitelli: Note - this is bad. If we can't have this API in EAP 6.x, it means we can't run the agent in the host controllers AFAIK.

josejulio: I could use reflection to call that method if exists. (when running on EAP-7)

pilhuhn: Mark it as TODO and continue
We need to 1st get this to work in standalone mode before we need to look at domain mode.
Also I believe that a EAP7 domain controller can control EAP6 managed servers, so that we could shoot for a 'hybrid' if all fails.

josejulio:
I got it to "work" (notice the quotes) on EAP6.

screenshot from 2016-11-29 00-35-54


Though is far from complete, the tests are broken, still need alternatives for methods and check the installer. Currently i have to modify the installed data, add jars and modify the xml.

pilhuhn:
On 29 Nov 2016, at 7:37, Josejulio Martínez wrote:

> I got it to "work" (notice the quotes) on EAP6.

Awesome!!!
:+1:

jmazzitelli:
Is there some way we can keep the original code intact and have this EAP6 backport supported and built from within, say, a separate branch or tag? Because once the EAP6 support goes away, we are going to want to return back to using only the EAP7 APIs.

jshaughn:
> Is there some way we can keep the original code intact...

Perhaps go so far as to have the code execute EAP6 APIs only when configured for EAP6 support (sysprop defaulting to true, or something like that)?  So that if there is an issue with EAP6 API and customers don't use EAP6, they could force it to EAP7.  And then when EAP6 support goes away we can easily deprecate and then get rid of the old support.

Maybe it's asking too much but just thought I'd put it out there.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
```diff
@@ -365,6 +357,7 @@ public void addDependencies(ServiceTarget target, ServiceBuilder<MonitorService>
 
         // bind the API to JNDI so other apps can use it, and prepare to build the binder service
         // Note that if we are running in host controller or similiar, JNDI binding is not available.
+        /*
```
jmazzitelli: This JNDI stuff is going to be needed when third party devs want to write their own integration with Hawkular Inventory and Metrics. So we can't just rip it out. We would need some kind of code here to say , "if EAP6, don't bind to jndi, if EAP7 bind to jndi". We might want to consider aborting agent startup if the jndi-name attribute in the agent subsystem configuration (in standalone.xml) is set but we are on EAP6

josejulio: Ok, will do that.

josejulio:
Something strange happened, 
When trying to start this with EAP7 I get a NoSuchMethodException on SimpleAttributeDefinitionBuilder.setAllowNull [here](https://github.com/josejulio/hawkular-agent/blob/31167c1dd49e41149600fadb7a6ae77a65a7ec3f/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/SubsystemAttributes.java#L33-L33)

Though, if i revert https://github.com/hawkular/hawkular-agent/pull/271/commits/0a95bfad3177234e9536fa7e4ca1db099e6aa7d7 it works well on both EAP7 and EAP6. Any though on this?

josejulio:
I have some of these exceptions on EAP6

```
10:49:13,002 ERROR [org.hawkular.agent.monitor.protocol.EndpointService] (Hawkular-WildFly-Agent-Scheduler-Metrics-2) HAWKMONITOR010049: Could not access resources of endpoint [DMREndpointService[Endpoint[Local]:[null]]]: org.hawkular.agent.monitor.protocol.ProtocolException: Unsuccessful fetching DMR attribute [BlockingFailureCount]
	at org.hawkular.agent.monitor.protocol.dmr.DMRDriver.fetchAttribute(DMRDriver.java:158)
	at org.hawkular.agent.monitor.protocol.EndpointService.measureMetrics(EndpointService.java:447)
	at org.hawkular.agent.monitor.scheduler.MetricsCollector.run(MetricsCollector.java:60)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [rt.jar:1.8.0_111]
	at java.util.concurrent.FutureTask.run(FutureTask.java:266) [rt.jar:1.8.0_111]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) [rt.jar:1.8.0_111]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617) [rt.jar:1.8.0_111]
	at java.lang.Thread.run(Thread.java:745) [rt.jar:1.8.0_111]
	at org.jboss.threads.JBossThread.run(JBossThread.java:122) [jboss-threads-2.1.2.Final-redhat-1.jar:2.1.2.Final-redhat-1]
Caused by: org.hawkular.dmr.api.OperationFailureException: Could not perform operation [read-attribute]: JBAS014792: Unknown attribute BlockingFailureCount
	at org.hawkular.dmr.api.OperationBuilder$OperationResult.assertSuccess(OperationBuilder.java:422)
	at org.hawkular.agent.monitor.protocol.dmr.DMRDriver.fetchAttribute(DMRDriver.java:155)
	... 8 more
```

BlockingFailureCount does not exists there, so i think that we should also make a template for EAP6, right?

pilhuhn:
I think we should try not to have 2 different agents + configs, but rather detection of where it runs and then expressions when a certain subsystem or attribute is applicable. Something like the following

```
<subsystem name="eap6 messaging" applies="eap6">
 ...
</subsystem>
<subsystem name="cool eap7 messaging" applies="eap7">
 ...
</subsystem>
<subsystem name="bla">  <!-- applies to both -->
  <property name="foo"  applies="eap6".../>
```

Unfortunately this is a bit tricky as we don't only need to consider the "base versions" but also potential layered products. JDG6 is internally using some of the EAP6 stuff, so 'applies="eap6"' needs also to be applied for JDG6. This potentially also means to have a negation. Either in the form
' applies="jdg6,!eap6"' or in 2 attributes 'applies' and 'unless' (or whatever we want to name them).

josejulio:
I'm looking at the hawkular-dmr-client and [this](https://github.com/hawkular/hawkular-agent/blob/026db452f7d17184e835b0c784f7a26c66ec7bd5/hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java#L532-L532) (ModelDescriptionConstants.RESOLVE_EXPRESSIONS) used [here](https://github.com/hawkular/hawkular-agent/blob/master/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/protocol/dmr/DMRDriver.java#L139-L139) doesn't seem to exist on jboss-as (first introduced [here](https://github.com/wildfly/wildfly-core/commit/69031d97d5286c7321c1317c36a664f1efb3997c)).

Would a proper replacement be to fail when resolveExpressions = true ? or WARN about it ?

jmazzitelli:
wow, EAP6 doesn't let clients resolve expressions? We use this a lot in the `<metric-dmr>` definitions (see the resolve-expressions attribute). I don't think we should fail. WARN at most.

I don't know what it will look like when collecting metrics on EAP6 when they are expressions and we ask to resolve them.

This EAP 6 agent is going to look ugly :)

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
josejulio: It seems that JNDI stuff can also run on EAP6, it was only missing the ImmediateManagedReference class, which didn't seem hard to implement and use when the other isn't found. I'll push those changes later.

josejulio:
I'm going to ask if there is other way to do it.

josejulio:
Should we set the dependencies as jboss-as ?
It was giving me some problems when trying to run the tests(it was using the jboss-as test suite with wildfly-controller, not sure why yet) and when trying to run on EAP7 (but it can be fixed with reflection), but the other way around seems fine.
If is fine I'll use Wildfly as dependencies for the project.

For supporting expression on EAP6 it would need to obtain the expression and then (if an expression is received) request to the server to resolve it..

josejulio:
@pilhuhn @jmazzitelli 

What target do you think I should use for compiling with? 
Setting target=jboss-as gives me some runtime errors on EAP7 that needs to be fixed with some reflection code. But runs fine on EAP6
Setting target=wildfly runs well on both, but hides on compile time some potential runtime errors on EAP6.
IMO we should compile with wildfly and maybe (not sure how to do it) run a 'compilation test' on jboss-as to ensure compability with that one.

josejulio:
Made some changes so wildfly libs are default.
 Using Wildfly API as default. Only use jboss-as API if can't find Wildfly API.

There is some pending things, like:
- Test JNDI stuff (not sure how to test it)
- Resolve expressions (other PR?)
- Configuration depending product type (see https://github.com/hawkular/hawkular-agent/pull/271#issuecomment-264622141) would be ok to have that in other commit, as not directly related to EAP6.

@jmazzitelli could you review and possibly decide if this belongs on another branch and/or how to follow up this.
// cc @pilhuhn @jshaughn

jmazzitelli:
JNDI testing - we have integration tests for this: see https://github.com/hawkular/hawkular-agent/blob/master/hawkular-wildfly-agent-itest-parent/hawkular-wildfly-agent-all-itests/hawkular-wildfly-agent-cmdgw-itest/src/test/java/org/hawkular/agent/test/HawkularWildFlyAgentContextITest.java

I would prefer this to be on a separate "backport-to-eap6" branch rather than in master.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/RemoteDMRAdd.java
```diff
@@ -40,9 +45,15 @@ private RemoteDMRAdd() {
     }
 
     @Override
+    protected void performRuntime(OperationContext context, ModelNode operation, ModelNode model,
```
jmazzitelli: can you pull up all these performRuntime methods you had to add in the individual classes into MonitorServiceAddStepHandler? That way, it will only be a single place to add this and not have to touch the subclasses for this.

josejulio: Sure thing, I'll do it.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/util/OperationContextExtension.java
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+package org.hawkular.agent.monitor.util;
+
+import org.jboss.as.controller.OperationContext;
+import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
+import org.jboss.dmr.ModelNode;
+
+public class OperationContextExtension {
```
jmazzitelli: Can you put all these compatibility static methods in the one WildflyCompabilityUtils.java and thus do not need to have a separate "OperationContextExtension" and "PathAddressExtension" class? This way, there is a single place for all the compatibility utilities. I don't see a need for these two extra "Extension" classes since they just have static methods.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/util/PathAddressExtension.java
```diff
@@ -0,0 +1,104 @@
...
+package org.hawkular.agent.monitor.util;
+
+import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.PathElement;
+
+/**
+ * PathAddress methods found on wildfly-core
+ * https://github.com/wildfly/wildfly-core/blob/588145a00ca5ed3beb0027f8f44be87db6689db3/controller/src/main/java/org/jboss/as/controller/PathAddress.java
+ */
+public class PathAddressExtension {
```
jmazzitelli: Can you put all these compatibility static methods in the one WildflyCompabilityUtils.java and thus do not need to have a separate "OperationContextExtension" and "PathAddressExtension" class? This way, there is a single place for all the compatibility utilities. I don't see a need for these two extra "Extension" classes since they just have static methods.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/util/OperationContextExtension.java
josejulio: Will also do that, totally makes sense to just have one of those.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/util/WildflyCompabilityUtils.java
```diff
@@ -0,0 +1,128 @@
...
+
+import org.jboss.as.controller.SubsystemRegistration;
+import org.jboss.as.naming.ImmediateManagedReferenceFactory;
+import org.jboss.as.naming.ManagedReference;
+import org.jboss.as.naming.ManagedReferenceFactory;
+import org.jboss.as.naming.ServiceBasedNamingStore;
+import org.jboss.as.naming.service.BinderService;
+import org.jboss.msc.inject.Injector;
+
+public class WildflyCompabilityUtils {
```
jmazzitelli: typo - the word "Compability" is misspelled - should be "Compatibility" (missing the "ti").

josejulio: whoops. Will fix that, thanks!

josejulio:
@jmazzitelli  Addressed all the comments and rebased to master.

JNDI testing - That runs on the context of Wildfly? I suppose I also need to run them on a jboss-as container.

Forgot to mention... I'm manually modifying the module on EAP6, including the following libs (the jar files and module.xml)
```
<resource-root path="jackson-core-2.7.4.jar"/>
<resource-root path="jackson-databind-2.7.4.jar"/>
<resource-root path="jackson-annotations-2.7.4.jar"/>
<resource-root path="javax.json-1.0.3.jar"/>

```
Is there any know way to only use them if not already provided by wildfly ?

josejulio:
If we are going to have it on a separate branch, could you please open that branch? I'll point this PR to that branch.

jmazzitelli:
> If we are going to have it on a separate branch, could you please open that branch? I'll point this PR to that branch.

https://github.com/hawkular/hawkular-agent/tree/backport-eap6

josejulio:
> If we are going to have it on a separate branch, could you please open that branch? I'll point this PR to that branch.
> https://github.com/hawkular/hawkular-agent/tree/backport-eap6

thank you, just changed the base branch.

josejulio:
I'm going to remove all those `log.warn` -> `log.warn(String.format` by specifying a base logger version (which should be compatible with both eap6 and eap7).
Is there anything else that i should do?

josejulio:
Rebased to master, fixed some missing calls and added required dependencies into the pom.xml and module.xml template.

@jmazzitelli Could you please take a look at the last 3 commits?

hawkular-wildfly-agent/pom.xml
```diff
@@ -113,6 +113,7 @@
     <dependency>
       <groupId>org.jboss.logging</groupId>
       <artifactId>jboss-logging</artifactId>
+      <version>3.1.4.GA-redhat-2</version>
```
jmazzitelli: This one is worrying because we are setting a hardcoded version but yet are saying it is "provided". So we are not guaranteed the agent is going to run with this version of logging (it will be provided with whatever the server has).

Why do we need to hardcode this specific redhat version? And what happens if we run the agent in a server that doesn't have this version?

jmazzitelli: Also, I don't think we can have community builds with -redhat artifacts. Are they even available? Need to ask if we can have -redhat artifacts in our community builds.

josejulio: The thing is that EAP6.4 comes with that version (3.1.4.GA-redhat-2) and I need to compile the code with that version because of debugf, logf, warnf, etcf signature.

On `3.1.4.GA-redhat-2` we have these signatures:
```java
  void debugf(String format, Object... params);
  void debugf(String format, Object param1);
```

But on `3.3.0` we have these signatures:
```java
void debugf(String format, Object... params);
void debugf(String format, Object param1);
void debugf(String format, int arg);
void debugf(String format, long arg);
```

among others.

If we don't specify a version, it will compile with 3.3.0.
In the case of [this](https://github.com/josejulio/hawkular-agent/blob/83514a58f030928d9019495a1a73b06255aca3f4/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/protocol/EndpointService.java#L86-L86) debug message it will expect the method `void debugf(String format, int arg);` (If we don't specify a version).

When running on EAP6.4 we get the following when starting:

```
22:15:26,698 ERROR [stderr] (Hawkular WildFly Agent Full Discovery Scan-platform-1) Exception in thread "Hawkular WildFly Agent Full Discovery Scan-platform-1" java.lang.NoSuchMethodError: org.hawkular.agent.monitor.log.MsgLogger.debugf(Ljava/lang/String;I)V
22:15:26,699 ERROR [stderr] (Hawkular WildFly Agent Full Discovery Scan-platform-1) 	at org.hawkular.agent.monitor.protocol.EndpointService$InventoryListenerSupport.fireResourcesAdded(EndpointService.java:86)
22:15:26,699 ERROR [stderr] (Hawkular WildFly Agent Full Discovery Scan-platform-1) 	at org.hawkular.agent.monitor.protocol.EndpointService$DiscoveryResults.discoveryFinished(EndpointService.java:151)
22:15:26,700 ERROR [stderr] (Hawkular WildFly Agent Full Discovery Scan-platform-1) 	at org.hawkular.agent.monitor.protocol.EndpointService$1.run(EndpointService.java:286)
22:15:26,700 ERROR [stderr] (Hawkular WildFly Agent Full Discovery Scan-platform-1) 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
22:15:26,700 ERROR [stderr] (Hawkular WildFly Agent Full Discovery Scan-platform-1) 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
22:15:26,700 ERROR [stderr] (Hawkular WildFly Agent Full Discovery Scan-platform-1) 	at java.lang.Thread.run(Thread.java:745)
22:15:26,700 ERROR [stderr] (Hawkular WildFly Agent Full Discovery Scan-platform-1) 	at org.jboss.threads.JBossThread.run(JBossThread.java:122)
```

In the other hand, if we compile with `3.1.4.GA-redhat-2` (or maybe even a lower version). It will attempt to use void `debugf(String format, Object param1);` which exists on both versions, thus working on both EAP6 and EAP7.

WRT to the -redhat version. I'm going to try to use a lower non-redhat version.

josejulio: Updated the code to use `3.1.4.GA` instead.

jmazzitelli: What happens to an agent that is compiled with logging 3.1.4.GA but run inside EAP7+ ? If that works also, then this fix would be OK (make sure you add a small XML comment in the pom to briefly explain why that hardcoded version is in there).

If that doesn't work (that is, if we can't compile with 3.1.4.GA and have it run in EAP7+) that is going to mean we can't do what @pilhuhn  suggests and have common code run on both (since logging is all throughout everything, it would be hard to fix this - unless we write some kind of our own logging proxy and put something in that compatibility utility class you wrote.).

josejulio: That also works, It's because EAP7 logging lib also has debugf(String format, Object param1); signature.
Added the explanation on the XML file.

hawkular-wildfly-agent/pom.xml
```diff
@@ -116,6 +116,11 @@
     <dependency>
       <groupId>org.jboss.logging</groupId>
       <artifactId>jboss-logging</artifactId>
+      <!-- We need to hardcode this version to ensure that calls like LOG.debugf("x", 1) don't link to
+      void debugf(String format, int arg);
+      Thus breaking compatibility with EAP6 which doesn't have said method.
+      Instead we use 3.1.4.GA to ensure it uses void debugf(String format, String param1); which is provided
```
jmazzitelli: I think you mean "Object param1", correct? (not String).

josejulio: Correct, just fixed it. Thanks for pointing out.

hawkular-wildfly-agent-feature-pack/src/main/resources/modules/system/add-ons/hawkular-agent/org/hawkular/agent/main/module.xml
```diff
@@ -72,6 +72,13 @@
     <artifact name="${httpmime}" />
      -->
 
+    <!-- EAP6 -->
+    <artifact name="${com.fasterxml.jackson.core:jackson-core}"/>
+    <artifact name="${com.fasterxml.jackson.core:jackson-databind}"/>
+    <artifact name="${com.fasterxml.jackson.core:jackson-annotations}"/>
+    <artifact name="${com.fasterxml.jackson.core:jackson-core}"/>
```
jmazzitelli: this is listed two times - remove one of them (jackson-core).

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
```diff
@@ -529,7 +530,18 @@ public T resolveExpressions() {
 
         @SuppressWarnings("unchecked")
         public T resolveExpressions(boolean resolveExpressions) {
-            baseNode.get(ModelDescriptionConstants.RESOLVE_EXPRESSIONS).set(resolveExpressions);
```
jmazzitelli, line 532: We should still set this because if this agent is also monitoring a remote EAP that happens to be EAP7 (which I think is still possible), this will be supported.

What we should do here is hardcode the string (don't use the constant, use the actual "resolve-expressions" string) but still do a check to see if the field exists so we can still log the warning. But the warning should say, "the local app server instance does not support resolving expressions").

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRemoveStepHandler.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
+ * Copyright 2015-2017 Red Hat, Inc. and/or its affiliates
```
jmazzitelli: since this file is no longer changed, can we revert it back (remove this change to the license header). One less file to be considered "changed" for this PR.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/SubsystemRemove.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Red Hat, Inc. and/or its affiliates
+ * Copyright 2015-2017 Red Hat, Inc. and/or its affiliates
```
jmazzitelli: since this file is no longer changed, can we revert it back (remove this change to the license header). One less file to be considered "changed" for this PR.

jmazzitelli:
@josejulio once those changes are made to address my last 4 comments, I would like for you to rebase and squash this PR into 1 commit. We are going to get ready to put this into master. First I am going to attempt to get two assemblies built from this one branch so we don't have to have multiple branches producing artifacts.

hawkular-wildfly-agent-feature-pack/src/main/resources/modules/system/add-ons/hawkular-agent/org/hawkular/agent/main/module.xml
josejulio: Yep, forgot to.

hawkular-dmr-client/src/main/java/org/hawkular/dmr/api/OperationBuilder.java
josejulio: Right.
I suppose that as a follow-up we could also call the resolve-expression command on the result of this operation if the target is EAP6

[standalone@localhost:29999 /] :resolve-expression(expression=

jmazzitelli:
OK, here is the plan. This PR will get merged into the backport-eap6 branch. I will then try to get EAP6-specific artifacts built for the wf-extension and feature-pack - that will be a PR on this backport-eap6 branch also.

After that is done (hopefully that works), I will then have to refactor the installer and h-services download servlet so it knows about these two different agents.

PR #276, opened 2016-12-02
Connection URL res config property is an expression

jmazzitelli:


jmazzitelli:
This addresses HWKAGENT-153

PR #280, opened 2017-01-23
feature pack and wf-extension built for both EAP6 and EAP7/Wildfly 10+

jmazzitelli:


jmazzitelli:
I think this is done. I just want to do some more manual testing before merging.

jmazzitelli:
Note that this does NOT alter the original metric-dmr and resource-type-dmr elements in the subsystem xml. We'll do that in another PR for this issue: https://issues.jboss.org/browse/HAWKULAR-1177

jmazzitelli:
@josejulio @pilhuhn I am going to merge this today unless there are objections (note, this will merge into backport-eap6 branch, NOT master - this only will go into master once I get the EAP6 metrics defined properly in the EAP6 agent - that will be a separate PR which I talk about below).

I have the installer able to install the agent into an EAP6. The download servlet is able to serve up the EAP6 agent extension (via ?appserver=eap6 query parameter) so this will go into h-services build once all of this is released (see https://github.com/hawkular/hawkular-services/pull/110) . I tested on EAP6 and it works (except for the metrics that need to be adjusted to match the resources that exist on EAP6 - that is a separate issue i will work on next. This means fixing the metric-dmr and resource-dmr to match the resources in EAP6. See https://issues.jboss.org/browse/HAWKULAR-1177).

@josejulio is working as a separate issue being able to run on EAP 7.1. But, again, that is independent of this work and will be done in a separate PR.

Let me know if you object to something in here.

PR #283, opened 2017-01-31
[WIP] HWKAGENT-156 - Support for EAP7.1

josejulio:
- Adds support for restarting hawkular-agent when modifying a property with the flag RESTART_RESOURCE_SERVICE.

Still need to:
- [x] Check support for EAP6
- [x] Workaround an exception (see MonitorService.java)

### DIFF
--- a/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java
+++ b/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java
@@ -18,7 +18,6 @@ package org.hawkular.agent.monitor.extension;
 
 import java.util.Collection;
 
-import org.hawkular.agent.monitor.service.MonitorService;
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
@@ -27,7 +26,6 @@ import org.jboss.as.controller.RestartParentWriteAttributeHandler;
 import org.jboss.dmr.ModelNode;
 import org.jboss.msc.service.ServiceController;
 import org.jboss.msc.service.ServiceName;
-import org.jboss.msc.service.ServiceRegistry;
 
 public class MonitorServiceRestartParentAttributeHandler extends RestartParentWriteAttributeHandler {
 
@@ -64,13 +62,5 @@ public class MonitorServiceRestartParentAttributeHandler extends RestartParentWr
     protected void recreateParentService(OperationContext context, PathAddress parentAddress, ModelNode parentModel)
             throws OperationFailedException {
         SubsystemAdd.INSTANCE.performRuntime(context, null, parentModel);
-        try {
-            ServiceName name = SubsystemExtension.SERVICE_NAME;
-            ServiceRegistry serviceRegistry = context.getServiceRegistry(true);
-            MonitorService service = (MonitorService) serviceRegistry.getRequiredService(name).getValue();
-            service.startMonitorService();
-        } catch (Exception e) {
-            throw new OperationFailedException("Failed to recreate Hawkular WildFly Agent service", e);
-        }
     }
 }
--- a/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java
+++ b/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java
@@ -23,6 +23,7 @@ import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
 import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.RestartParentWriteAttributeHandler;
+import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
 import org.jboss.dmr.ModelNode;
 import org.jboss.msc.service.ServiceController;
 import org.jboss.msc.service.ServiceName;
@@ -61,6 +62,8 @@ public class MonitorServiceRestartParentAttributeHandler extends RestartParentWr
     @Override
     protected void recreateParentService(OperationContext context, PathAddress parentAddress, ModelNode parentModel)
             throws OperationFailedException {
-        SubsystemAdd.INSTANCE.performRuntime(context, null, parentModel);
+        SubsystemAdd.INSTANCE.performRuntime(context,
+                new ModelNode().set(ModelDescriptionConstants.OP_ADDR, parentAddress.toModelNode()),
+                parentModel);
     }
 }
--- a/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java
+++ b/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2015-2017 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.hawkular.agent.monitor.extension;
+
+import java.util.Collection;
+
+import org.hawkular.dmr.api.OperationBuilder;
+import org.jboss.as.controller.AttributeDefinition;
+import org.jboss.as.controller.OperationContext;
+import org.jboss.as.controller.OperationFailedException;
+import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.RestartParentWriteAttributeHandler;
+import org.jboss.dmr.ModelNode;
+import org.jboss.msc.service.ServiceController;
+import org.jboss.msc.service.ServiceName;
+
+public class MonitorServiceRestartParentAttributeHandler extends RestartParentWriteAttributeHandler {
+
+    public MonitorServiceRestartParentAttributeHandler(AttributeDefinition... definitions) {
+        super("subsystem", definitions);
+    }
+
+    public MonitorServiceRestartParentAttributeHandler(Collection<AttributeDefinition> definitions) {
+        super("subsystem", definitions);
+    }
+
+    @Override
+    protected ServiceName getParentServiceName(PathAddress parentAddress) {
+        return SubsystemExtension.SERVICE_NAME;
+    }
+
+    @Override
+    protected boolean isResourceServiceRestartAllowed(OperationContext context, ServiceController<?> service) {
+        return true;
+    }
+
+    @Override
+    protected boolean requiresRuntime(OperationContext context) {
+        return true;
+    }
+
+    @Override
+    protected void removeServices(OperationContext context, ServiceName parentService, ModelNode parentModel) throws OperationFailedException {
+        SubsystemRemove.INSTANCE.performRuntime(context, null, parentModel);
+    }
+
+    @Override
+    protected void recreateParentService(OperationContext context, PathAddress parentAddress, ModelNode parentModel) throws OperationFailedException {
+        SubsystemAdd.INSTANCE.performRuntime(context, null, parentModel);
+        ModelNode startOperation = OperationBuilder.byName("start")
+                .address(parentAddress)
+                .attribute("restart", "true")
+                .build();
+        new OperationSubsystemStart().execute(context, startOperation);
+    }
+}
--- a/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java
+++ b/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java
@@ -18,7 +18,7 @@ package org.hawkular.agent.monitor.extension;
 
 import java.util.Collection;
 
-import org.hawkular.dmr.api.OperationBuilder;
+import org.hawkular.agent.monitor.service.MonitorService;
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
@@ -27,6 +27,7 @@ import org.jboss.as.controller.RestartParentWriteAttributeHandler;
 import org.jboss.dmr.ModelNode;
 import org.jboss.msc.service.ServiceController;
 import org.jboss.msc.service.ServiceName;
+import org.jboss.msc.service.ServiceRegistry;
 
 public class MonitorServiceRestartParentAttributeHandler extends RestartParentWriteAttributeHandler {
 
@@ -54,17 +55,22 @@ public class MonitorServiceRestartParentAttributeHandler extends RestartParentWr
     }
 
     @Override
-    protected void removeServices(OperationContext context, ServiceName parentService, ModelNode parentModel) throws OperationFailedException {
+    protected void removeServices(OperationContext context, ServiceName parentService, ModelNode parentModel)
+            throws OperationFailedException {
         SubsystemRemove.INSTANCE.performRuntime(context, null, parentModel);
     }
 
     @Override
-    protected void recreateParentService(OperationContext context, PathAddress parentAddress, ModelNode parentModel) throws OperationFailedException {
+    protected void recreateParentService(OperationContext context, PathAddress parentAddress, ModelNode parentModel)
+            throws OperationFailedException {
         SubsystemAdd.INSTANCE.performRuntime(context, null, parentModel);
-        ModelNode startOperation = OperationBuilder.byName("start")
-                .address(parentAddress)
-                .attribute("restart", "true")
-                .build();
-        new OperationSubsystemStart().execute(context, startOperation);
+        try {
+            ServiceName name = SubsystemExtension.SERVICE_NAME;
+            ServiceRegistry serviceRegistry = context.getServiceRegistry(true);
+            MonitorService service = (MonitorService) serviceRegistry.getRequiredService(name).getValue();
+            service.startMonitorService();
+        } catch (Exception e) {
+            throw new OperationFailedException("Failed to recreate Hawkular WildFly Agent service", e);
+        }
     }
 }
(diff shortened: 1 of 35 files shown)


hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override
+    protected void recreateParentService(OperationContext context, PathAddress parentAddress, ModelNode parentModel) throws OperationFailedException {
+        SubsystemAdd.INSTANCE.performRuntime(context, null, parentModel);
+        ModelNode startOperation = OperationBuilder.byName("start")
+                .address(parentAddress)
+                .attribute("restart", "true")
+                .build();
+        new OperationSubsystemStart().execute(context, startOperation);
```
josejulio: When modifying an attribute marked with Flag.RESTART_RESOURCE_SERVICES  this immediately recreates the subsystem (removes, add, start).

I haven't found yet a way to mark locally this service as *needs reload* or something similar. I suppose we could also implement a way to mark it and a reload operation to execute it on-demand.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java
```diff
@@ -376,6 +376,50 @@ public void addDependencies(ServiceTarget target, ServiceBuilder<MonitorService>
         // Note that if we are running in host controller or similiar, JNDI binding is not available.
         String jndiName = bootConfiguration.getApiJndi();
         boolean bindJndi = (jndiName == null || jndiName.isEmpty() || processType.isManagedDomain()) ? false : true;
+        /**
+         * Currently, when trying to re-bind the jndi it triggers... Need to work this out when everything else is done.
+         12:59:09,467 ERROR [org.jboss.as.controller.management-operation] (management-handler-thread - 2) WFLYCTL0013: Operation ("write-attribute") failed - address: ([("subsystem" => "hawkular-wildfly-agent")]): org.jboss.msc.service.DuplicateServiceException: Service jboss.naming.context.java.global.hawkular.agent.api is already registered
```
josejulio: When below code is executed (specifically `binderBuilder.install();`) it throws an exception.
I suppose we could unbind this when calling the remove operation or verify here if is already installed before trying to install.

jmazzitelli:
Question: suppose someone changes N attributes (for example, suppose I want to change the collection interval for 5 metric sets defined in the subsystem) - is this going to restart the agent repeatedly for N times? Or does the system somehow avoid this?

josejulio:
If executed all of them in a single compose operation, it will restart once.
If executed each one, it will restart once for every attribute with Flag.RESTART_RESOURCE_SERVICE.

josejulio:
An alternative for this behavior would be to define our own operation to reload the service and mark a reload required on the subsystem. 
I'm not aware of any way to actually mark a service for reload but we could also implement something to do that.

As far i could tell, on EAP7, these properties acted like RESTART_ALL_SERVICES. See AbstractWriteAttributeHandler [1] and ReloadRequiredWriteAttributeHandler [2] for more info.

[1] https://github.com/wildfly/wildfly-core/blob/8814b2af45a23af556a58a685bc0f0a02a71dca4/controller/src/main/java/org/jboss/as/controller/AbstractWriteAttributeHandler.java#L104-L106
[2] https://github.com/wildfly/wildfly-core/blob/73cc07034d5ca660a6bab890e2e951b80c6f34d9/controller/src/main/java/org/jboss/as/controller/ReloadRequiredWriteAttributeHandler.java#L54-L55

jmazzitelli:
I think we need to fix OperationStart. All that does is start or restart MonitorService but it doesn't get that service to reload the new configuration. If you changed some agent config via CLI, just merely starting/stopping the agent via startMonitorService is not enough because it would still have the old configuration still.

We need to somehow get OperationStart to invoke SubsystemRemove (to stop) and SubsystemAdd (to start). But Wildfly has to trigger those calls because Wildfly needs to pass in the proper context and config to those methods. But we don't want to remote the entire agent subsystem's config - we just want to force WildFly to restart our services.

What we have to end up doing is creating a new MonitorService object, passing in the new config to the constructor and then having the start() called. this is what SubsystemAdd does.

jmazzitelli:
Ignore my last comment. The fix would be to change the new method: org.hawkular.agent.monitor.extension.MonitorServiceRestartParentAttributeHandler.recreateParentService()

I'm experimenting with this as the new implementation of that method:

```
    @Override
    protected void recreateParentService(OperationContext context, PathAddress parentAddress, ModelNode parentModel)
            throws OperationFailedException {
        SubsystemAdd.INSTANCE.performRuntime(context, null, parentModel);
        try {
            ServiceName name = SubsystemExtension.SERVICE_NAME;
            ServiceRegistry serviceRegistry = context.getServiceRegistry(true);
            MonitorService service = (MonitorService) serviceRegistry.getRequiredService(name).getValue();
            service.startMonitorService();
        } catch (Exception e) {
            throw new OperationFailedException("Failed to recreate Hawkular WildFly Agent service", e);
        }
    }
```

josejulio:
Merged josejulio/hawkular-agent#1 and rebased.

jmazzitelli:
The travis failure happens to me locally as well. Investigating...

jmazzitelli:
OK, the itest error is definitely showing a bug in the new PR code. I do not think the implementation of `MonitorServiceRestartParentAttributeHandler.recreateParentService()` is correct.

Here's the stack trace of importance:

```
07:28:17,235 ERROR [org.jboss.as.controller.management-operation] (management-handler-thread - 1) WFLYCTL0013: Operation ("write-attribute") failed - address: ([
    ("subsystem" => "hawkular-wildfly-agent"),
    ("managed-servers" => "default"),
    ("remote-jmx" => "Remote JMX")
]): java.lang.IllegalArgumentException: Missing storage adapter configuration: {"enabled" : false, "url" : "http://localhost:8080/jolokia-war", "username" : null, "password" : null, "security-realm" : null, "set-avail-on-shutdown" : null, "resource-type-sets" : "MainJMX,MemoryPoolJMX", "tenant-id" : null, "metric-id-template" : null, "metric-tags" : "feed=%FeedId,Label Remote=Value Remote"}
	at org.hawkular.agent.monitor.extension.MonitorServiceConfigurationBuilder.determineStorageAdapterConfig(MonitorServiceConfigurationBuilder.java:857)
	at org.hawkular.agent.monitor.extension.MonitorServiceConfigurationBuilder.<init>(MonitorServiceConfigurationBuilder.java:94)
	at org.hawkular.agent.monitor.extension.SubsystemAdd.performRuntime(SubsystemAdd.java:46)
	at org.hawkular.agent.monitor.extension.MonitorServiceRestartParentAttributeHandler.recreateParentService(MonitorServiceRestartParentAttributeHandler.java:66)
	at org.jboss.as.controller.RestartParentWriteAttributeHandler.applyUpdateToRuntime(RestartParentWriteAttributeHandler.java:72)
	at org.jboss.as.controller.AbstractWriteAttributeHandler$1.execute(AbstractWriteAttributeHandler.java:104)
	at org.jboss.as.controller.AbstractOperationContext.executeStep(AbstractOperationContext.java:890)
	at org.jboss.as.controller.AbstractOperationContext.processStages(AbstractOperationContext.java:659)
	at org.jboss.as.controller.AbstractOperationContext.executeOperation(AbstractOperationContext.java:370)
	at org.jboss.as.controller.OperationContextImpl.executeOperation(OperationContextImpl.java:1344)
	at org.jboss.as.controller.ModelControllerImpl.internalExecute(ModelControllerImpl.java:392)
	at org.jboss.as.controller.ModelControllerImpl.execute(ModelControllerImpl.java:217)
```

So we are writing an attribute metric-tags to the remote JMX managed server element (remote-jmx) so the MonitorServiceRestartParentAttributeHandler.recreateParentService is getting the ModelNode of that resource, but it is expecting the FULL agent configuration model node starting at the top level subsystem.

The itest is writing an attribute when it gets this error message - here's the test error message:

```
FAILED: testDmrResources
org.hawkular.dmrclient.FailureException: Cannot set metric tags on remote JMX managed server: {
    "outcome" => "failed",
    "failure-description" => "WFLYCTL0158: Operation handler failed: java.lang.IllegalArgumentException: Missing storage adapter configuration: {\"enabled\" : false, \"url\" : \"http://localhost:8080/jolokia-war\", \"username\" : null, \"password\" : null, \"security-realm\" : null, \"set-avail-on-shutdown\" : null, \"resource-type-sets\" : \"MainJMX,MemoryPoolJMX\", \"tenant-id\" : null, \"metric-id-template\" : null, \"metric-tags\" : \"feed=%FeedId,Label Remote=Value Remote\"}",
    "rolled-back" => true
}
	at org.hawkular.agent.jmx.test.LocalAndRemoteJmxITest.setMetricTagsOnJmxManagedServers(LocalAndRemoteJmxITest.java:228)
	at org.hawkular.agent.jmx.test.LocalAndRemoteJmxITest.testDmrResources(LocalAndRemoteJmxITest.java:51)
```

The test code that is writing the new metric-tags attribute is here: 

https://github.com/hawkular/hawkular-agent/blob/master/hawkular-wildfly-agent-itest-parent/hawkular-wildfly-agent-all-itests/hawkular-wildfly-agent-jmx-itest/src/test/java/org/hawkular/agent/jmx/test/LocalAndRemoteJmxITest.java#L212-L236

So this needs to be fixed before we can merge this.

jmazzitelli:
OK, the easier way to replicate this problem (rather than running that long itest) is to simply run the agent and via the JBoss CLI execute this command:

```
jboss-cli.sh --connect '--commands=/subsystem=hawkular-wildfly-agent/managed-servers=default/local-dmr=Local/:write-attribute(name=metric-tags,value="a=b")'
```

You will get this error which is essentially what the itest failure is:

`
{
    "outcome" => "failed",
    "failure-description" => "WFLYCTL0158: Operation handler failed: java.lang.IllegalArgumentException: Missing storage adapter configuration: {\"enabled\" : true, \"set-avail-on-shutdown\" : null, \"resource-type-sets\" : \"Standalone Environment,Deployment,Web Component,EJB,Datasource,XA Datasource,JDBC Driver,Transaction Manager,Messaging,Socket Binding Group,Clustering,Hawkular\", \"tenant-id\" : null, \"metric-id-template\" : null, \"metric-tags\" : \"a=b\"}",
    "rolled-back" => true
}
`

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java
```diff
@@ -0,0 +1,76 @@
...
+    @Override
+    protected void removeServices(OperationContext context, ServiceName parentService, ModelNode parentModel)
+            throws OperationFailedException {
+        SubsystemRemove.INSTANCE.performRuntime(context, null, parentModel);
+    }
+
+    @Override
+    protected void recreateParentService(OperationContext context, PathAddress parentAddress, ModelNode parentModel)
+            throws OperationFailedException {
+        SubsystemAdd.INSTANCE.performRuntime(context, null, parentModel);
```
jmazzitelli: The problem here is "parentModel" is the model of the parent resource whose attribute is changing. But this code expects that ModelNode to be of the FULL agent (starting at the top level &lt;subsystem> element). Is there a way we can walk up the tree and get that top level mode but still keep the new changes that are in parentModel?

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/protocol/EndpointService.java
```diff
@@ -80,33 +82,49 @@
...
+        private final ReadWriteLock inventoryListenerRWLock = new ReentrantReadWriteLock();
 
         public void fireResourcesAdded(List<Resource<L>> resources) {
             if (!resources.isEmpty()) {
-                LOG.debugf("Firing inventory event for [%d] added/modified resources", resources.size());
-                InventoryEvent<L> event = new InventoryEvent<L>(EndpointService.this, resources);
-                for (InventoryListener inventoryListener : inventoryListeners) {
-                    inventoryListener.resourcesAdded(event);
+                try {
+                    inventoryListenerRWLock.readLock().lock();
```
jmazzitelli: acquisition of all locks should always come outside (just before) entering the try-catch. Because if something goes wrong, the finally block is going to try to unlock a lock you have not acquired. So swap lines 89-90. Do the same for the others.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/protocol/EndpointService.java
```diff
@@ -80,33 +82,49 @@
...
         }
 
         public void fireResourcesRemoved(List<Resource<L>> resources) {
             if (!resources.isEmpty()) {
-                LOG.debugf("Firing inventory event for [%d] removed resources", resources.size());
-                InventoryEvent<L> event = new InventoryEvent<L>(EndpointService.this, resources);
-                for (InventoryListener inventoryListener : inventoryListeners) {
-                    inventoryListener.resourcesRemoved(event);
+                try {
+                    inventoryListenerRWLock.readLock().lock();
```
jmazzitelli: acquisition of all locks should always come outside (just before) entering the try-catch. Because if something goes wrong, the finally block is going to try to unlock a lock you have not acquired. So swap lines 104-105. Do the same for the others.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/protocol/EndpointService.java
```diff
@@ -80,33 +82,49 @@
...
         }
 
         public void fireDiscoveryComplete() {
-            LOG.debugf("Firing inventory event for discovery complete");
-            DiscoveryEvent<L> event = new DiscoveryEvent<L>(EndpointService.this, getResourceManager(),
-                    getResourceTypeManager());
-            for (InventoryListener inventoryListener : inventoryListeners) {
-                inventoryListener.discoveryCompleted(event);
+            try {
+                inventoryListenerRWLock.readLock().lock();
```
jmazzitelli: acquisition of all locks should always come outside (just before) entering the try-catch. Because if something goes wrong, the finally block is going to try to unlock a lock you have not acquired. So swap lines 118-119. Do the same for the others.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/protocol/EndpointService.java
```diff
@@ -226,9 +244,14 @@ public ProtocolDiagnostics getDiagnostics() {
      * @param listener to add
      */
     public void addInventoryListener(InventoryListener listener) {
-        status.assertInitialOrStopped(getClass(), "addInventoryListener()");
-        this.inventoryListenerSupport.inventoryListeners.add(listener);
-        LOG.debugf("Added inventory listener [%s] for endpoint [%s]", listener, getMonitoredEndpoint());
+        try {
+            this.inventoryListenerSupport.inventoryListenerRWLock.writeLock().lock();
```
jmazzitelli: acquisition of all locks should always come outside (just before) entering the try-catch. Because if something goes wrong, the finally block is going to try to unlock a lock you have not acquired. So swap lines 247-248. Do the same for the others.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/protocol/EndpointService.java
```diff
@@ -237,9 +260,14 @@ public void addInventoryListener(InventoryListener listener) {
      * @param listener to remove
      */
     public void removeInventoryListener(InventoryListener listener) {
-        status.assertInitialOrStopped(getClass(), "removeInventoryListener()");
-        this.inventoryListenerSupport.inventoryListeners.remove(listener);
-        LOG.debugf("Removed inventory listener [%s] for endpoint [%s]", listener, getMonitoredEndpoint());
+        try {
+            this.inventoryListenerSupport.inventoryListenerRWLock.writeLock().lock();
```
jmazzitelli: acquisition of all locks should always come outside (just before) entering the try-catch. Because if something goes wrong, the finally block is going to try to unlock a lock you have not acquired. So swap lines 263-264. Do the same for the others.

jmazzitelli (changes requested):

jmazzitelli:
I think we have a major problem. itests show failure trying to start up in domain mode. i have a feeling the stuff we added to plug into the servers lifecycle doesn't work in domain mode. Still investigating.

jmazzitelli:
This is definitely a problem. Install the agent now in a host controller, and try to start it via domain.sh and it immediately aborts startup.

Install:

```
java -jar target/hawkular-wildfly-agent-installer-full-*.jar --target-location=$HOME/bin/eap/jboss-eap-7.0 --username=jdoe --password=password --server-url=http://localhost:8080 --module-dist=classpath: --target-config=$HOME/bin/eap/jboss-eap-7.0/domain/configuration/host.xml
```

Run `$HOME/bin/eap/jboss-eap-7.0/bin/domain.sh`

See error:

```
05:50:30,799 INFO  [org.jboss.modules] (main) JBoss Modules version 1.5.1.Final-redhat-1
05:50:30,983 INFO  [org.jboss.as.process.Host Controller.status] (main) WFLYPC0018: Starting process 'Host Controller'
[Host Controller] 05:50:31,617 INFO  [org.jboss.modules] (main) JBoss Modules version 1.5.1.Final-redhat-1
[Host Controller] 05:50:31,864 INFO  [org.jboss.msc] (main) JBoss MSC version 1.2.6.Final-redhat-1
[Host Controller] 05:50:31,930 INFO  [org.jboss.as] (MSC service thread 1-7) WFLYSRV0049: JBoss EAP 7.0.0.GA (WildFly Core 2.1.2.Final-redhat-1) starting
[Host Controller] 05:50:33,390 INFO  [org.jboss.as.host.controller] (Controller Boot Thread) WFLYHC0003: Creating http management service using network interface (management) port (9990) securePort (-1)
[Host Controller] 05:50:33,395 INFO  [org.xnio] (MSC service thread 1-4) XNIO version 3.3.6.Final-redhat-1
[Host Controller] 05:50:33,408 INFO  [org.xnio.nio] (MSC service thread 1-4) XNIO NIO Implementation Version 3.3.6.Final-redhat-1
[Host Controller] 05:50:33,418 ERROR [org.jboss.as.controller.management-operation] (Controller Boot Thread) WFLYCTL0013: Operation ("add") failed - address: ([
[Host Controller]     ("host" => "master"),
[Host Controller]     ("subsystem" => "hawkular-wildfly-agent")
[Host Controller] ]) - failure description: "WFLYCTL0216: Management resource '[(\"subsystem\" => \"hawkular-wildfly-agent\")]' not found"
[Host Controller] 05:50:33,456 FATAL [org.jboss.as.host.controller] (Controller Boot Thread) WFLYHC0034: Host Controller boot has failed in an unrecoverable manner; exiting. See previous messages for details.
[Host Controller] 05:50:33,473 INFO  [org.jboss.as] (MSC service thread 1-7) WFLYSRV0050: JBoss EAP 7.0.0.GA (WildFly Core 2.1.2.Final-redhat-1) stopped in 7ms
[Host Controller] 
```

jmazzitelli:
found the problem - in SubsystemAdd this subsystem address must be incorrect when running in domain mode and thus this throws an exception and halts the boot up:

```
        ModelNode subsystemConfig = Resource.Tools.readModel(context.readResourceFromRoot(
                WildflyCompatibilityUtils.parseCLIStyleAddress("/subsystem=" + SubsystemExtension.SUBSYSTEM_NAME)
        ));
```

jmazzitelli (approved):

jmazzitelli:
Finally got travis to go green. All other testing looks good. This is ready to be merged.

PR #284, opened 2017-02-02
HWKAGENT-157 abort first startup if being asked to start up again

jmazzitelli:


jmazzitelli:
@josejulio please peer review this when you can.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/OperationSubsystemStart.java
```diff
@@ -44,7 +44,17 @@ public void execute(OperationContext opContext, ModelNode model) throws Operatio
...
             }
             if (!service.isMonitorServiceStarted()) {
-                service.startMonitorService();
+                Thread newThread = new Thread(new Runnable() {
+                    @Override
+                    public void run() {
+                        try {
+                            service.startMonitorService();
+                        } catch (Throwable t) {
+                        }
```
josejulio: Shouldn't we log anything bad that happens here?

jmazzitelli: This reason why I do not log anything here is because startMonitorService() will already log that error - any Throwable thrown from method startMonitorService is first logged by that method, see here:

https://github.com/jmazzitelli/hawkular-agent/blob/fb0e9878953ffc2bbd2922277af7f154e6568176/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/service/MonitorService.java#L646

You can see startMonitorService() has a large try-catch that surrounds the entire start up code and the catch block logs the exception. Thus to avoid simply duplicating that error log message, I didn't log that exception a second time here.

josejulio: But those Throwables are already catched by `startMonitorService` and as far as i can see they aren't re-throw.
There wasn't any try/catch surrounding `service.startMonitorService()` before, why do we need one now?

jmazzitelli:
good point. I didn't realize I don't even need the try-catch. It has been removed.

josejulio:
LGTM

PR #286, opened 2017-02-06
HAWKULAR-1183 immutable flag

jmazzitelli:


eap6-support/hawkular-wildfly-agent-feature-pack-eap6/src/main/resources/subsystem-templates/hawkular-wildfly-agent.xml
```diff
@@ -622,11 +623,11 @@
         <operation-dmr name="Shutdown" internal-name="shutdown" > <!-- shutdown and restart -->
```
jkremser: Shouldn't be those `{start,restart,stop}-servers`, `restart`, `reload`, `stop` operations marked also with the modifies=true? Does it make sense to stop/start a server in a container instead of stopping, starting the whole container or pod?

jmazzitelli: I don't know. You tell me :) My thinking for not setting the modifies=true on those is that this isn't modifying anything (in the sense there is no persistent changes being made). We are just starting/stopping processes, but nothing is "changing" (where "change" is, say, a modification to a property stored in standalone.xml, as one example).

jkremser: I don't know either, @pilhuhn tell us :]

I see your logic, those operations don't modify anything. The question is whether we want them enabled in the containers. According to that pet & cattle analogy, the containers should be shot in the leg = not restarting the processes in it.

jmazzitelli:
This is ready to be merged.

eap6-support/hawkular-wildfly-agent-feature-pack-eap6/src/main/resources/subsystem-templates/hawkular-wildfly-agent.xml
jmazzitelli: @theute also said he thinks the power ops should be disabled (see comment in https://issues.jboss.org/browse/HAWKULAR-1183) so I will do that before merging this.

jmazzitelli:
Just made a change as per discussion in the JIRA - all power ops will be disabled when immutable.

PR #287, opened 2017-02-07
Try to read machine-id from property if set by user.

pilhuhn:
This is a proposal to read the machine id from a property if set there.
That can then in container deployments be set to the container-id 
(in fact in this case it could just be the hostname = server name = feed id)

See also https://github.com/pilhuhn/hawkular-agent/blob/hawkfly/docker-dist/src/main/resources/standalone.conf#L58


jmazzitelli:
This looks good. I will merge this but first I want to get PR #283 in first (that also modifies Util - let's wait so we don't churn that PR).

PR #288, opened 2017-02-07
HWKAGENT-159 - support ${x} in server url and add download-server-url opt

jmazzitelli:


jmazzitelli:
@pilhuhn you might want to check this first before I merge. Make sure it does what you want.

You still must supply the three parts to the URL (protocol, host, port) but you can use ${x} now. So this is valid:

`--server-url=http://${HAWKULAR_SERVER_HOST}:${HAWKULAR_SERVER_PORT:8080}`

You can even use ${HAWKULAR_PROTO} for the protocol

`--server-url=${HAWKULAR_PROTO}://${HAWKULAR_SERVER_HOST}:${HAWKULAR_SERVER_PORT:8080}`

but if you do, you have to remember to pass in --keystore-path and all the rest if you plan on being able to have it use https.

pilhuhn:
Close, but not yet. I have

`--server-url 'http://${env.HAWKULAR_SERVER_ADDR}:${env.HAWKULAR_SERVER_PORT}'`

Which should end up as  `<storage-adapter url="http://${env.HAWKULAR_SERVER_ADDR}:${env.HAWKULAR_SERVER_PORT}">`, telling the server to populate the url from the 2 env vars.

I just got the thought to wonder if we should ditch this and instead allow to tell the installer "don't modify standalone.xml' and then explicitly provide a hand-crafted one. Disadvantage is though that we need yet another place to modify if agent bits change.

pilhuhn:
But then - perhaps I have the wrong installer bits

jmazzitelli:
You don't need the url attribute. You will see that will put those values in the outbound socket binding and it works. I tested that. You can put ${x} values in the socket bindings and it works.

Make sure you got the latest builds and try again. This is something I tested and confirmed it worked.

pilhuhn:
Works now. I forgot to also compile in ...installer-full, so always ended up with the bits prior to this change.
Thanks Mazz!

PR #290, opened 2017-02-09
HAWKULAR-1153 Initial support enable/disable/update collection intervals

jshaughn:
- Add command to handle new cmdgw request/response for updating metric collection intervals

@jmazzitelli One more for you. Please review, merge needs to wait for update commons dep, though.

### DIFF
--- a/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/cmd/UpdateCollectionIntervalsCommand.java
+++ b/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/cmd/UpdateCollectionIntervalsCommand.java
@@ -1,0 +1,214 @@
+/*
+ * Copyright 2015-2017 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.hawkular.agent.monitor.cmd;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.hawkular.agent.monitor.extension.MonitorServiceConfiguration.AbstractEndpointConfiguration;
+import org.hawkular.agent.monitor.inventory.MonitoredEndpoint;
+import org.hawkular.agent.monitor.log.AgentLoggers;
+import org.hawkular.agent.monitor.log.MsgLogger;
+import org.hawkular.agent.monitor.protocol.EndpointService;
+import org.hawkular.agent.monitor.protocol.dmr.DMRNodeLocation;
+import org.hawkular.agent.monitor.protocol.dmr.DMRSession;
+import org.hawkular.agent.monitor.util.WildflyCompatibilityUtils;
+import org.hawkular.bus.common.BasicMessageWithExtraData;
+import org.hawkular.bus.common.BinaryData;
+import org.hawkular.cmdgw.api.UpdateCollectionIntervalsRequest;
+import org.hawkular.cmdgw.api.UpdateCollectionIntervalsResponse;
+import org.hawkular.dmr.api.DmrApiException;
+import org.hawkular.dmr.api.OperationBuilder;
+import org.hawkular.dmr.api.OperationBuilder.CompositeOperationBuilder;
+import org.hawkular.dmr.api.OperationBuilder.OperationResult;
+import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.client.ModelControllerClient;
+
+/**
+ * Update the specified metric and avail type collection intervals.  Performs a bulk update to prevent
+ * excessive agent restarts. Because metric types are not guaranteed to be consistent across agents, it is
+ * not a failure if a requested metric type does not exist.
+ *
+ * @author Jay Shaughnessy
+ */
+public class UpdateCollectionIntervalsCommand
+        extends AbstractResourcePathCommand<UpdateCollectionIntervalsRequest, UpdateCollectionIntervalsResponse> {
+    private static final MsgLogger log = AgentLoggers.getLogger(UpdateCollectionIntervalsCommand.class);
+    public static final Class<UpdateCollectionIntervalsRequest> REQUEST_CLASS = UpdateCollectionIntervalsRequest.class;
+
+    public UpdateCollectionIntervalsCommand() {
+        super("Update Collection Intervals", "DMR Nodes");
+    }
+
+    /** @see org.hawkular.agent.monitor.cmd.AbstractResourcePathCommand#createResponse() */
+    @Override
+    protected UpdateCollectionIntervalsResponse createResponse() {
+        return new UpdateCollectionIntervalsResponse();
+    }
+
+    @Override
+    protected BinaryData execute(
+            ModelControllerClient mcc,
+            EndpointService<DMRNodeLocation, DMRSession> endpointService,
+            String modelNodePath,
+            BasicMessageWithExtraData<UpdateCollectionIntervalsRequest> envelope,
+            UpdateCollectionIntervalsResponse response,
+            CommandContext context,
+            DMRSession dmrContext) throws Exception {
+
+        PathAddress agentAddress = WildflyCompatibilityUtils.parseCLIStyleAddress(modelNodePath);
+
+        UpdateCollectionIntervalsRequest request = envelope.getBasicMessage();
+        Map<PathAddress, String> metricTypes = filterUpdates(mcc, request.getMetricTypes(), false, agentAddress);
+        Map<PathAddress, String> availTypes = filterUpdates(mcc, request.getAvailTypes(), true, agentAddress);
+
+        if (isEmpty(metricTypes) && isEmpty(availTypes)) {
+            log.debug("Skipping collection interval update, no valid type updates provided.");
+            return null;
+        }
+
+        CompositeOperationBuilder<?> composite = OperationBuilder.composite();
+        addUpdates(composite, metricTypes);
+        addUpdates(composite, availTypes);
+        // Don't let the restart services hook kick in, instead do a deferred restart
+        composite.allowResourceServiceRestart(false);
+        // Now refresh the monitor service with the updated config. Delay the restart to allow time for the
+        // command response to be returned.
+        composite.byNameOperation("start")
+                .address(agentAddress)
+                .attribute("refresh", "true")
+                .attribute("delay", "500") //  enough time to get the response sent back
+                .parentBuilder();
+
+        try {
+            OperationResult<?> opResult = composite.execute(mcc).assertSuccess();
+            setServerRefreshIndicator(opResult, response);
+
+        } catch (DmrApiException e) {
+            log.errorf("Failed to update collection intervals: %s", e.getMessage());
+            throw e;
+        }
+
+        return null;
+    }
+
+    private Map<PathAddress, String> filterUpdates(ModelControllerClient mcc, Map<String, String> updates,
+            boolean isAvail, PathAddress agentAddress) {
+
+        if (isEmpty(updates)) {
+            return null;
+        }
+
+        String agentPath = agentAddress.toCLIStyleString();
+        Map<PathAddress, String> result = new HashMap<>();
+
+        for (Map.Entry<String, String> entry : updates.entrySet()) {
+            String typeId = entry.getKey();
+            String interval = entry.getValue();
+
+            PathAddress metricType = null;
+            try {
+                String path = isAvail ? toAvailPath(typeId, agentPath) : toMetricPath(typeId, agentPath);
+                metricType = WildflyCompatibilityUtils.parseCLIStyleAddress(path);
+            } catch (Exception e1) {
+                log.warnf("Unable to update interval for invalid type [%s]", typeId);
+                continue;
+            }
+
+            try {
+                Integer.valueOf(interval);
+            } catch (Exception e) {
+                log.warnf("Unable to update interval for type [%s], invalid interval [%s]", typeId, interval);
+                continue;
+            }
+
+            if (!isValidAddress(mcc, metricType)) {
+                log.warnf("Skipping collection interval update on invalid metric type [%s]", typeId);
+                continue;
+            }
+
+            result.put(metricType, interval);
+        }
+
+        return result;
+    }
+
+    private boolean isValidAddress(ModelControllerClient mcc, PathAddress address) {
+        try {
+            OperationBuilder.readAttribute()
+                    .address(address)
+                    .name("interval")
+                    .includeDefaults()
+                    .execute(mcc)
+                    .assertSuccess();
+            return true;
+
+        } catch (Exception e) {
+            return false;
+        }
+    }
+
+    private void addUpdates(CompositeOperationBuilder<?> composite, Map<PathAddress, String> updates) {
+        if (!isEmpty(updates)) {
+            for (Map.Entry<PathAddress, String> entry : updates.entrySet()) {
+                PathAddress metricType = entry.getKey();
+                String interval = entry.getValue();
+
+                composite.writeAttribute()
+                        .address(metricType)
+                        .attribute("interval", interval)
+                        .allowResourceServiceRestart(false)
+                        .parentBuilder();
+                composite.writeAttribute()
+                        .address(metricType)
+                        .attribute("time-units", "seconds")
+                        .allowResourceServiceRestart(false)
+                        .parentBuilder();
+            }
+        }
+    }
+
+    private String toMetricPath(String metricTypeId, String agentPath) {
+        String[] a = metricTypeId.split("~");
+        if (a.length != 2) {
+            throw new IllegalArgumentException("MetricTypeId must be of form MetricTypeSet~MetricTypeName");
+        }
+        return agentPath + (agentPath.endsWith("/") ? "" : "/") + "metric-set-dmr=" + a[0] + "/metric-dmr=" + a[1];
+    }
+
+    private String toAvailPath(String availTypeId, String agentPath) {
+        String[] a = availTypeId.split("~");
+        if (a.length != 2) {
+            throw new IllegalArgumentException("AvailTypeId must be of form AvailTypeSet~AvailTypeName");
+        }
+        return agentPath + (agentPath.endsWith("/") ? "" : "/") + "avail-set-dmr=" + a[0] + "/avail-dmr=" + a[1];
+    }
+
+    private boolean isEmpty(Map<?, ?> c) {
+        return null == c || c.isEmpty();
+    }
+
+    @Override
+    protected void validate(BasicMessageWithExtraData<UpdateCollectionIntervalsRequest> envelope,
+            MonitoredEndpoint<? extends AbstractEndpointConfiguration> endpoint) {
+    }
+
+    @Override
+    protected void validate(String modelNodePath,
+            BasicMessageWithExtraData<UpdateCollectionIntervalsRequest> envelope) {
+    }
+}
--- a/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java
+++ b/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java
@@ -16,6 +16,9 @@
  */
 package org.hawkular.agent.monitor.extension;
 
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ALLOW_RESOURCE_SERVICE_RESTART;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OPERATION_HEADERS;
+
 import java.util.Collection;
 
 import org.hawkular.agent.monitor.service.MonitorService;
@@ -34,11 +37,11 @@ import org.jboss.msc.service.ServiceRegistry;
 
 public class MonitorServiceRestartParentAttributeHandler extends RestartParentWriteAttributeHandler {
 
-    private static final OperationContext.AttachmentKey<Integer> RECREATE_COUNTER =
-            OperationContext.AttachmentKey.create(Integer.class);
+    private static final OperationContext.AttachmentKey<Integer> RECREATE_COUNTER = OperationContext.AttachmentKey
+            .create(Integer.class);
 
-    private static final OperationContext.AttachmentKey<Integer> REMOVE_COUNTER =
-            OperationContext.AttachmentKey.create(Integer.class);
+    private static final OperationContext.AttachmentKey<Integer> REMOVE_COUNTER = OperationContext.AttachmentKey
+            .create(Integer.class);
 
     public MonitorServiceRestartParentAttributeHandler(AttributeDefinition... definitions) {
         super("subsystem", definitions);
@@ -51,11 +54,6 @@ public class MonitorServiceRestartParentAttributeHandler extends RestartParentWr
     @Override
     protected ServiceName getParentServiceName(PathAddress parentAddress) {
         return SubsystemExtension.SERVICE_NAME;
-    }
-
-    @Override
-    protected boolean isResourceServiceRestartAllowed(OperationContext context, ServiceController<?> service) {
-        return true;
     }
 
     @Override
@@ -95,12 +93,20 @@ public class MonitorServiceRestartParentAttributeHandler extends RestartParentWr
                             new ModelNode().set(ModelDescriptionConstants.OP_ADDR, parentAddress.toModelNode()),
                             parentModel);
                 } else if (counter < 0) {
-                    throw new OperationFailedException("The recreateParentService step got added more times than needed - " +
-                            "This shouldn't happen.");
+                    throw new OperationFailedException(
+                            "The recreateParentService step got added more times than needed - " +
+                                    "This shouldn't happen.");
                 }
                 WildflyCompatibilityUtils.operationContextStepCompleted(context);
             }
         }, OperationContext.Stage.RUNTIME);
+    }
+
+    // By default we always allow a restart
+    @Override
+    protected boolean isResourceServiceRestartAllowed(final OperationContext context,
+            final ServiceController<?> service) {
+        return true;
     }
 
     @Override
@@ -116,8 +122,33 @@ public class MonitorServiceRestartParentAttributeHandler extends RestartParentWr
             }
         }
 
-        return super.applyUpdateToRuntime(context, operation, attributeName, resolvedValue, currentValue,
-                handbackHolder);
+        // When updating agent config we by default allow immediate restarts, the caller does not have to be aware
+        // of which attributes may or may not force the restart.  In some situations, namely updates made by
+        // an agent command like UpdateCollectionIntervalsCommand, we need to defer the restart because otherwise
+        // the command response will never be sent back.  To defer the restart the operation must explicitly
+        // have the header ALLOW_RESOURCE_SERVICE_RESTART=false (note, each operation in a composite must assign
+        // it individually). In this situation not only do we want to defer the restart, we want to avoid setting
+        // the server to 'reload-required'.  To do this we temporarily mark the parent as restarted, and we must
+        // also return false from this method.
+        ModelNode headers = operation.has(OPERATION_HEADERS) ? operation.get(OPERATION_HEADERS) : null;
+        boolean restartAllowedOptionSet = (headers != null) && headers.hasDefined(ALLOW_RESOURCE_SERVICE_RESTART);
+        boolean restartAllowed = !restartAllowedOptionSet || headers.get(ALLOW_RESOURCE_SERVICE_RESTART).asBoolean();
+        Object tempOwner = null;
+        PathAddress address = null;
+        try {
+            if (!restartAllowed) {
+                address = getParentAddress(WildflyCompatibilityUtils.getCurrentAddress(context, operation));
+                tempOwner = new Object();
+                context.markResourceRestarted(address, tempOwner);
+            }
+            super.applyUpdateToRuntime(context, operation, attributeName, resolvedValue, currentValue, handbackHolder);
+
+        } finally {
+            if (!restartAllowed && null != address && null != tempOwner) {
+                context.revertResourceRestarted(address, tempOwner);
+            }
+        }
+        return false;
     }
 
     protected MonitorService getMonitorService(OperationContext opContext) {
--- a/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/OperationSubsystemStart.java
+++ b/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/OperationSubsystemStart.java
@@ -16,10 +16,16 @@
  */
 package org.hawkular.agent.monitor.extension;
 
+import java.util.concurrent.atomic.AtomicReference;
+
 import org.hawkular.agent.monitor.service.MonitorService;
+import org.hawkular.agent.monitor.service.ServiceStatus;
+import org.hawkular.agent.monitor.util.WildflyCompatibilityUtils;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
 import org.jboss.as.controller.OperationStepHandler;
+import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.registry.Resource;
 import org.jboss.dmr.ModelNode;
 import org.jboss.logging.Logger;
 import org.jboss.msc.service.ServiceName;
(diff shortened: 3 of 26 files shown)


hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/cmd/UpdateCollectionIntervalsCommand.java
```diff
@@ -0,0 +1,148 @@
...
+import org.jboss.as.controller.client.ModelControllerClient;
+
+/**
+ * Update the specified metric and avail type collection intervals.  Performs a bulk update to prevent
+ * excessive agent restarts. Because metric types are not guaranteed to be consistent across agents, it is
+ * not a failure if a requested metric type does not exist.
+ *
+ * @author Jay Shaughnessy
+ */
+public class UpdateCollectionIntervalsCommand
```
jmazzitelli, line 48: Right now this is picking up the superclass' modifiesResource method which returns true. See:

https://github.com/hawkular/hawkular-agent/blob/master/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/cmd/AbstractResourcePathCommand.java#L267

This means if the agent is configured as "immutable" clients will not be able to update collection intervals.

@pilhuhn @theute what is the thoughts on that? If the agent is immutable (i.e. running in a container), do we still want to be able to change the collection intervals (which will change agent configuration)?

With this code right now you canNOT change collection intervals when the agent is immutable.

If we want to be able to change collection intervals, we need this class to implement its own modifiesResource() to simply return false.

theute: Of course I should have seen this before I commented on the JIRA: https://issues.jboss.org/browse/HAWKULAR-1157

I have mixed feelings... mostly because immutable == read-only so far, and now if that triggers a reload, the server will be non-responsive for a short period of time. (and then the guy who monitor the app, may have a negative impact on the server)

jmazzitelli: @theute I think there is a misunderstanding about what this new reload feature we introduced yesterday is doing. When an agent configuration setting changes now, ONLY THE AGENT SERVICE will restart. The entire app server does not restart - it should still be responsive to requests. It only affects the agent responsiveness (the agent won't be able to respond to websocket requests from the CFME UI, for example; and it will temporarily stop collecting metrics - but the restart really only takes a few seconds to complete).

theute: Thanks for the explanation. That definitely helps. 
@jdoyleoss Do you think we should allow metrics enablement/disablement and collection interval change operations for immutable containers ? This would ultimately change the configuration file of the container and not the docker container image itself. See also mazz's comment about the small impact. (Note that this is preparation work for the UX part we need to achieve later)

jdoyleoss: I don't think we should allow it.  The change could go away at any time, new instances won't have it. I think it will just lead to confusion. If a user needs to change it in order to do some debugging, they can shell into the box and make the change.

jshaughn:
@jmazzitelli I removed the DNM.  Note that in addition to adding the new command, this changes the behavior of the  RESTART_RESOURCE_SERVICES when updating attributes.  It now means that the caller must explicitly allow the immediate restart, otherwise it will be deferred.  This is important for a command updating the agent because if the agent restarts immediately the command response is lost.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/OperationSubsystemStart.java
```diff
@@ -37,22 +37,34 @@ public void execute(OperationContext opContext, ModelNode model) throws Operatio
...
+            long delay = model.get("delay").asLong(0L);
+
+            Thread newThread = new Thread(new Runnable() {
+                @Override
+                public void run() {
+                    if (delay > 0) {
+                        try {
+                            Thread.sleep(delay);
+                        } catch (Exception e) {
+                            // just continue with start
```
jmazzitelli: This needs to abort and not continue. What is probably happening here is the sleep got an InterruptedException, which usually means the app server has been sent a shutdown signal and all threads are being told to shutdown. So I would re-thrown the exception here which helps the app server shutdown.

jmazzitelli: Actually, just return, don't throw exception - don't know if that default exception handler will puke in the log file. just silently return here if the sleep is interrupted.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/OperationSubsystemStart.java
```diff
@@ -31,28 +33,43 @@
...
-                newThread.start();
-            }
+                }
+            }, "Hawkular WildFly Agent Operation Start Thread"));
+            newThread.get().setDaemon(true);
+            newThread.get().start();
+
         } catch (ServiceNotFoundException snfe) {
             throw new OperationFailedException("Cannot restart Hawkular Monitor service - it is disabled", snfe);
         } catch (Exception e) {
```
jmazzitelli, line 88: you put the newThread interrupt in the wrong spot. You want a catch (InterruptedException ie) here just above the catch (Exception e). In that new catch clause is where you call newThread.get().interrupt(). That interrupt will (potentially) interrupt the sleep and when interrupting the sleep, you just simply "return". You want to tell the inner thread to stop doing what its doing - and you do taht by catching the interrupt and just returning.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/OperationSubsystemStart.java
```diff
@@ -31,28 +33,43 @@
...
+            long delay = model.get("delay").asLong(0L);
+
+            newThread.set(new Thread(new Runnable() {
+                @Override
+                public void run() {
+                    if (delay > 0) {
+                        try {
+                            Thread.sleep(delay);
+                        } catch (Exception e) {
+                            newThread.get().interrupt();
```
jmazzitelli: see my comment below - this should simply be "return;" here in the catch clause. If the sleep is interruptted, you don't want it to interrupt itself and keep going! You want it to return and which effectively kills the thread.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/OperationSubsystemStart.java
jshaughn: yes, duh me. I'm tired.

jmazzitelli:
OK, LGTM. Once green, I'll merge and get a release out

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java
```diff
@@ -54,11 +53,6 @@ protected ServiceName getParentServiceName(PathAddress parentAddress) {
     }
 
     @Override
-    protected boolean isResourceServiceRestartAllowed(OperationContext context, ServiceController<?> service) {
```
josejulio, line 57: I just would like to note that if we don't allow a restart (i.e. `isResourceServiceRestartAllowed` returns `false`) the whole server will be put in `reload required´ state. See [here](https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/RestartParentWriteAttributeHandler.java#L77-L78) and [here](https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/AbstractWriteAttributeHandler.java#L155-L156) for more info.

This can be avoided by always returning false on `applyUpdateToRuntime`, but then, we wouldn't know if the agent needs a restart.

jmazzitelli: Yes, correct. And for our websocket clients, they will be told about the new state of the server and if they need to tell it to restart or reload via the restart flag in the json response (me and Jay were reminded of this fact yesterday when the itests failed due to this flag being set when we weren't expecting it to be :)

DMR clients will also know this because they, too, are given the reload/restart flag in the response.

jshaughn: It is correct but I'm not sure it's acceptable.  It may be OK in the interim but with the PR as it stands if you execute the UpdateCollectionIntervalsCommand it leaves the server in a 'reload-required' state.  That's not really good because in fact the server does not need a reload, we are restarting the agent and that is all that should be required.  I think we can get around this, I'll look at it this morning.  Because no one is calling UpdateCollectionIntervalsCommand at the moment, I don't think it needs to block a merge.

josejulio: I'm not sure if we could replicate what the `reload-required` does:
Mark the agent in a `restart-required` state and notify in every operation against the agent that a restart is required.

jmazzitelli: I don't think we have thought about this fully yet :) I am starting to think we want @josejulio 's code to be the default behavior and somehow get your stuff to workaround that @jshaughn .

I think we want (by default) any change to attributes (or composite operations to change multiple ones) to restart the agent service.

If @jshaughn 's use-case does not want that, we need to figure out how to accomodate that.

jshaughn: OK, I just updated the PR, the key change is to have `MonitorServiceRestartParentAttributeHandler.applyUpdateToRuntime` always return `false`.  By doing this a config update can basically allow or not allow restart.  Either way the update will never set the server to 'reload-required'.  So, by default we don't restart, only if you explicitly add the option to allow restarts.  If you still feel we need to allow restart by default then we'll need some more work here.  Let me know what you think.

josejulio: Maybe we could try (in a separate PR) to somehow mark the agent as `needs-restart` to let know the user that we are dealing with an agent running using old configurations.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/cmd/UpdateCollectionIntervalsCommand.java
```diff
@@ -0,0 +1,200 @@
...
+            return null;
+        }
+
+        CompositeOperationBuilder<?> composite = OperationBuilder.composite();
+        addUpdates(composite, metricTypes);
+        addUpdates(composite, availTypes);
+        // Don't let the restart hooks kick in, instead do a deferred restart
+        //composite.allowResourceServiceRestart();
+        composite.byNameOperation("start")
+                .address(PathAddress.parseCLIStyleAddress("/subsystem=hawkular-wildfly-agent/"))
```
josejulio: This won't work on EAP6, please to use this: [WildflyCompatibilityUtils.parseCLIStyleAddress](https://github.com/josejulio/hawkular-agent/blob/b873a144a7b2bf6c8e081ffcd583f23b55a1c515/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/util/WildflyCompatibilityUtils.java#L119-L119)

`WildflyCompatibilityUtils.parseCLIStyleAddress("/subsystem=hawkular-wildfly-agent/")`

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/cmd/UpdateCollectionIntervalsCommand.java
```diff
@@ -0,0 +1,200 @@
...
+        Map<PathAddress, String> result = new HashMap<>();
+
+        for (Map.Entry<String, String> entry : updates.entrySet()) {
+            String typeId = entry.getKey();
+            String interval = entry.getValue();
+
+            PathAddress metricType = null;
+            try {
+                String path = isAvail ? toAvailPath(typeId) : toMetricPath(typeId);
+                metricType = PathAddress.parseCLIStyleAddress(path);
```
josejulio: This won't work on EAP6, please to use this: [WildflyCompatibilityUtils.parseCLIStyleAddress](https://github.com/josejulio/hawkular-agent/blob/b873a144a7b2bf6c8e081ffcd583f23b55a1c515/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/util/WildflyCompatibilityUtils.java#L119-L119)

`WildflyCompatibilityUtils.parseCLIStyleAddress(path)`

jmazzitelli:
@jshaughn there are breakages in here when running in EAP6 - see @josejulio comments

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java
jshaughn: OK, the latest revision changes it around such that restarts kick in automatically and are deferred only if the operation explicitly disallows the restart in a header.  So basically, the new UpdateCollectionsIntervalCommand defers the restart, anything else would restart automatically. Also, I updated with the EAP6 fix, thanks @josejulio .

jshaughn:
@jmazzitelli , @josejulio , unless you have more comments I think this PR is ready for merge.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java
```diff
@@ -116,8 +122,29 @@ protected boolean applyUpdateToRuntime(OperationContext context, ModelNode opera
...
+        // it individually). In this situation not only do we want to defer the restart, we want to avoid setting
+        // the server to 'reload-required'.  To do this we temporarily mark the parent as restarted, and we must
+        // also return false from this method.
+        ModelNode headers = operation.has(OPERATION_HEADERS) ? operation.get(OPERATION_HEADERS) : null;
+        boolean restartAllowedOptionSet = (headers != null) && headers.hasDefined(ALLOW_RESOURCE_SERVICE_RESTART);
+        boolean restartAllowed = !restartAllowedOptionSet || headers.get(ALLOW_RESOURCE_SERVICE_RESTART).asBoolean();
+        Object tempOwner = null;
+        PathAddress address = null;
+        if (!restartAllowed) {
+            address = getParentAddress(context.getCurrentAddress());
```
josejulio: Sorry missed this one before, we don't have this method on EAP6 `OperationContext.getCurrentAddress`
Please use [this](https://github.com/hawkular/hawkular-agent/blob/master/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/util/WildflyCompatibilityUtils.java#L195) one instead.
`WildflyCompatibilityUtils.getCurrentAddress(context, operation)`
Note that on EAP6 the address is taken from the operation.

josejulio:
@jshaughn I left inline a comment about EAP6 support.
Also, do you know if we should also override [revertUpdateToRuntime](https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/RestartParentWriteAttributeHandler.java#L157)?

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java
jshaughn: Thanks, I updated this.

jshaughn:
> Also, do you know if we should also override revertUpdateToRuntime?

This is a good question.  I don't think so but looking back at the code, I think it will be safer if I move the 
`revertResourceRestarted` into a finally, to ensure we don't block things on some failure. I'll update the code now...

jmazzitelli:
@josejulio what do you think about this PR? Do you think it is ready? Have you looked at EAP6 compatibility?

josejulio:
It looks good. I can test it on EAP6 to verify the compatibility.

jshaughn:
@jmazzitelli I added a new ImmutableITest under cmdgw tests, it basically makes an agent immutable, which kicks off a restart, and then tries to do a collection interval update, which fails appropriately.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/cmd/UpdateCollectionIntervalsCommand.java
```diff
@@ -0,0 +1,209 @@
...
+            log.debug("Skipping collection interval update, no valid type updates provided.");
+            return null;
+        }
+
+        CompositeOperationBuilder<?> composite = OperationBuilder.composite();
+        addUpdates(composite, metricTypes);
+        addUpdates(composite, availTypes);
+        // Don't let the restart hooks kick in, instead do a deferred restart
+        composite.allowResourceServiceRestart(false);
+        composite.byNameOperation("start")
```
josejulio: Umh, sorry I didn't notice this before.
I suppose you want to load the new values, however i don't think a restart would do that.
A restart doesn't re-load the configuration, you will need to remove [1] and add [2] the subsystem for that to happen [3].

[1] https://github.com/josejulio/hawkular-agent/blob/ee7c30c4658472c6178da90fec0102371b5b0574/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java#L73-L73
[2] https://github.com/josejulio/hawkular-agent/blob/ee7c30c4658472c6178da90fec0102371b5b0574/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java#L92-L92
[3] https://github.com/josejulio/hawkular-agent/blob/b874396edf35f28bb1fc07a7ab4f9165457ab29c/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/SubsystemAdd.java#L70-L70

jshaughn: whoah, OK, so the restarts are only restarting the monitor service but not really boostrapping the subsystem.  Certainly we need to config to re-read.  Let me look into this now. Thanks.

jmazzitelli:
@jshaughn is going to look into implementing a new "restartWithNewConfiguration" method in MonitorService. Our OperationSubsystemStart can call that. Hopefully, that gives some new feature he can use.

jshaughn:
Right,  so @josejulio had a very good catch there, unfortunately we couldn't just do a remove/add of the subsystem because that would return us to the previous issue of killing the agent before the agent command had a chance to send it's command response.  That's why we wanted to use the start operation (with a short delay).  The latest commit enhances the start operation with a new `refresh` attribute, which is like `restart` but using the latest config for the new runtimeConfiguration.  You guys can take a look, @jmazzitelli, I roughly used the approach we came up with.   I couldn't quite use the command you put together, but finally arrived at something similar that worked.

Unfortunately there is still some issue in the cmdgw itests, looking...

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/OperationSubsystemStart.java
```diff
@@ -31,35 +36,70 @@
...
+                        if (delay > 0) {
+                            Thread.sleep(delay);
+                        }
+
+                        MonitorServiceConfiguration runtimeConfig = null;
+
+                        if (restart) {
+                            if (refresh) {
+                                PathAddress address = WildflyCompatibilityUtils.parseCLIStyleAddress(
+                                        "/subsystem=hawkular-wildfly-agent/");
```
jmazzitelli: you can't do this - hard-coding the address. The address is different when running in domain mode (there is a parent whose name is not known at compile time).

josejulio:
Just tested this on EAP6.4.
Is working good :+1:

jshaughn:
@jmazzitelli @josejulio Thanks for the reviews and help.  I am hopeful this is now in a final state.

jmazzitelli:
@josejulio one more time :) see if your EAP6 stuff works.

josejulio:
EAP6 stuff is working good! :+1:

jmazzitelli:
I tried running itests locally and got a failure. I think there is still timing issues here. I see this:

```
09:45:37,819 ERROR [org.hawkular.agent.monitor.protocol.EndpointService] (Hawkular WildFly Agent Full Discovery Scan-Local-1) HAWKMONITOR010049: Could not access resources of endpoint [DMREndpointService[Endpoint[Local]:[null]]]: java.lang.IllegalStateException: [org.hawkular.agent.monitor.protocol.dmr.DMREndpointService] must be in state [RUNNING] rather than [STOPPED] to perform [discoverChildren()]
	at org.hawkular.agent.monitor.service.ServiceStatus.assertRunning(ServiceStatus.java:34)
	at org.hawkular.agent.monitor.protocol.EndpointService.discoverChildren(EndpointService.java:339)
	at org.hawkular.agent.monitor.protocol.EndpointService.access$500(EndpointService.java:80)
	at org.hawkular.agent.monitor.protocol.EndpointService$1.run(EndpointService.java:303)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
```

The test failure is reported as:

```
FAILED: testAddDeployment
java.lang.AssertionError: Expected: to match pattern "\QDeployApplicationResponse={\"destinationFileName\":\"hawkular-wildfly-agent-helloworld-war.war\",\"resourcePath\":\"/t;itest-rest-tenant/f;d5091b6f-e3ee-4a30-a122-518ebf73bfaf/r;Local~~\",\"destinationSessionId\":\"{{sessionId}}\",\"status\":\"OK\",\"message\":\"Performed [Deploy] on a [Application] given by Inventory path [/t;itest-rest-tenant/f;d5091b6f-e3ee-4a30-a122-518ebf73bfaf/r;Local~~]\"}\E.*"
     but: was <DeployApplicationResponse={"resourcePath":"/t;itest-rest-tenant/f;d5091b6f-e3ee-4a30-a122-518ebf73bfaf/r;Local~~","destinationSessionId":"fhaevMySZNpah_E44hB0yWdymnMlWrL5a-1Gzndb","status":"ERROR","message":"Could not perform [Deploy] on a [Application] given by inventory path [/t;itest-rest-tenant/f;d5091b6f-e3ee-4a30-a122-518ebf73bfaf/r;Local~~] requested on [2017-02-18T14:45Z]: java.lang.IllegalArgumentException: Cannot deploy application: unknown resource [/t;itest-rest-tenant/f;d5091b6f-e3ee-4a30-a122-518ebf73bfaf/r;Local~~]"}>
	at org.hawkular.cmdgw.ws.test.TestWebSocketClient$ExpectedEvent$ExpectedMessage.match(TestWebSocketClient.java:525)
	at org.hawkular.cmdgw.ws.test.TestWebSocketClient$TestListener.handle(TestWebSocketClient.java:971)
	at org.hawkular.cmdgw.ws.test.TestWebSocketClient$TestListener.onMessage(TestWebSocketClient.java:998)
	at okhttp3.internal.ws.RealWebSocket$1.onMessage(RealWebSocket.java:62)
	at okhttp3.internal.ws.WebSocketReader.readMessageFrame(WebSocketReader.java:242)
```

jmazzitelli:
I think the waitForAgent stuff isn't right - or even the OperationSubsystemStatus isn't going to always be correct. The agent's `isMonitorServiceStarted` returns the `started` boolean flag. However, that is not set to false until the full shutdown code is finished (that is, until `stopMonitorService` finishes).

So, if you call `isMonitoredServiceStarted` in the middle of the agent shutting down, it will still return "true" and make you think the service is started (i.e. you don't know its in the middle of a shutdown).

Here's what I see in the logs and what I think it means (I chopped some log messages that I didn't think pertinent, but the logs you see here are in the order in which they were printed):

```
FINE: Received message[2] from WebSocket: [ActualMessage [type=application/vnd.okhttp.websocket+text; charset=utf-8, body=UpdateCollectionIntervalsResponse={"resourcePath":"/t;itest-rest-tenant/f;d5091b6f-e3ee-4a30-a122-518ebf73bfaf/r;Local~~/r;Local~%2Fsubsystem%3Dhawkular-wildfly-agent","destinationSessionId":"UhGkKU9w1hNOWpO2d-JotFueD-NLU-c6OJCSCQMY","status":"OK","message":"Performed [Update Collection Intervals] on a [DMR Nodes] given by Inventory path [/t;itest-rest-tenant/f;d5091b6f-e3ee-4a30-a122-518ebf73bfaf/r;Local~~/r;Local~%2Fsubsystem%3Dhawkular-wildfly-agent]"}]]
Feb 18, 2017 9:45:31 AM org.hawkular.cmdgw.ws.test.TestWebSocketClient$TestListener handle
FINE: Message[2]: [PASSED]
```

At this point, the UpdateCollectionIntervals test has passed....

```
Feb 18, 2017 9:45:31 AM org.hawkular.wildfly.agent.itest.util.AbstractITest waitForAgent
INFO: Checking [/subsystem=hawkular-wildfly-agent] status...
```

Now I think the test (UpdateCollectionIntervals) is waiting for the agent to restart (notice: we have not received any log messages about the agent restarting yet - that is coming next).

```
09:45:32,012 WARN  [org.hawkular.agent.monitor.extension.OperationSubsystemStart] (Hawkular WildFly Agent Operation Start Thread) Stopping Hawkular Monitor Service now, refresh requested.
09:45:32,013 INFO  [org.hawkular.agent.monitor.service.MonitorService] (Hawkular WildFly Agent Operation Start Thread) HAWKMONITOR010001: Stopping Hawkular WildFly Agent service
09:45:32,022 INFO  [org.hawkular.cmdgw.command.ws.server.FeedWebSocket] (default task-56) HAWKFEEDCOMM000015: WebSocket session [d5091b6f-e3ee-4a30-a122-518ebf73bfaf] of [/feed/{feedId}] closed. Reason=[CloseReason[1000,Shutting down FeedCommProcessor]]
09:45:32,028 INFO  [org.hawkular.agent.monitor.cmd.FeedCommProcessor] (OkHttp http://localhost.localdomain:8080/hawkular/command-gateway/feed/d5091b6f-e3ee-4a30-a122-518ebf73bfaf WebSocket Close Reply) HAWKMONITOR010035: Closed feed WebSocket connection to endpoint [ws://localhost.localdomain:8080/hawkular/command-gateway/feed/d5091b6f-e3ee-4a30-a122-518ebf73bfaf]. Code=[1000], Reason=[Shutting down FeedCommProcessor]
```

OK, the agent has shutdown as it is in the process of restarting.

```
Feb 18, 2017 9:45:36 AM org.hawkular.wildfly.agent.itest.util.AbstractITest waitForAgent
INFO: Agent [/subsystem=hawkular-wildfly-agent] status=STARTED, continuing...
```

The UpdateCollectionIntervals test PASSED its waitForAgent check. But it should not have. The agent really hasn't started yet - it just is in the middle (but has not finished) its shutdown. So the agent is reporting "started" but it really is about to be shutdown in a moment.

```
[Invoker 1785397234] Invoking @AfterMethod AbstractITest.after()[pri:0, instance:org.hawkular.agent.ws.test.UpdateCollectionIntervalsCommandITest@5f16132a]
[Invoker 1785397234] Invoking @BeforeMethod AbstractITest.before()[pri:0, instance:org.hawkular.agent.ws.test.StandaloneDeployApplicationITest@45ca843]
```

The "AfterMethod" for the UpdateCollectionIntervals test is run - this means that test is finished. Now notice the StandaloneDeployApplication test is about to run. But the agent is currently in the middle of shutting down right now!!! It hasn't even finished the restart yet.

```
[Invoker 1785397234] Invoking org.hawkular.agent.ws.test.StandaloneDeployApplicationITest.testAddDeployment
Got code 200 and message [OK] retries: http://127.0.0.1:8080/hawkular/inventory/traversal/f;d5091b6f-e3ee-4a30-a122-518ebf73bfaf/type=r
Got after 1 retries: http://127.0.0.1:8080/hawkular/inventory/traversal/f;d5091b6f-e3ee-4a30-a122-518ebf73bfaf/type=r
09:45:36,890 INFO  [org.hawkular.cmdgw.command.ws.server.UIClientWebSocket] (default task-58) HAWKFEEDCOMM000013: WebSocket session [fhaevMySZNpah_E44hB0yWdymnMlWrL5a-1Gzndb] opened for endpoint [/ui/ws]
Feb 18, 2017 9:45:36 AM org.hawkular.cmdgw.ws.test.TestWebSocketClient$TestListener onOpen
FINE: WebSocket opened
Feb 18, 2017 9:45:36 AM org.hawkular.cmdgw.ws.test.TestWebSocketClient$TestListener handle
FINE: Received message[0] from WebSocket: [ActualMessage [type=application/vnd.okhttp.websocket+text; charset=utf-8, body=WelcomeResponse={"sessionId":"fhaevMySZNpah_E44hB0yWdymnMlWrL5a-1Gzndb"}]]
Feb 18, 2017 9:45:36 AM org.hawkular.cmdgw.ws.test.TestWebSocketClient$TestListener handle
FINE: Message[0]: [PASSED]
Feb 18, 2017 9:45:36 AM org.hawkular.cmdgw.ws.test.TestWebSocketClient$MessageAnswer$1 run
FINE: Sending over WebSocket: DeployApplicationRequest={"authentication":{"username":"itest-rest","password":"3bd28cc7-450b-4b53-9b59-b74733497180"}, "resourcePath":"/t;itest-rest-tenant/f;d5091b6f-e3ee-4a30-a122-518ebf73bfaf/r;Local~~","destinationFileName":"hawkular-wildfly-agent-helloworld-war.war"}
Feb 18, 2017 9:45:36 AM org.hawkular.cmdgw.ws.test.TestWebSocketClient$MessageAnswer$1 run
FINE: No sleep configured
09:45:36,941 INFO  [org.hawkular.cmdgw.command.ws.server.AbstractGatewayWebSocket] (default task-60) HAWKFEEDCOMM000021: Received message [org.hawkular.cmdgw.api.DeployApplicationRequest] with binary data from WebSocket session [fhaevMySZNpah_E44hB0yWdymnMlWrL5a-1Gzndb] of [/ui/ws]
Feb 18, 2017 9:45:36 AM org.hawkular.cmdgw.ws.test.TestWebSocketClient$TestListener handle
FINE: Received message[1] from WebSocket: [ActualMessage [type=application/vnd.okhttp.websocket+text; charset=utf-8, body=GenericSuccessResponse={"message":"The request has been forwarded to feed [d5091b6f-e3ee-4a30-a122-518ebf73bfaf] (MessageId=ID:e8de9dca-f5e8-11e6-aa86-7db894f89301)"}]]
Feb 18, 2017 9:45:36 AM org.hawkular.cmdgw.ws.test.TestWebSocketClient$TestListener handle
FINE: Message[1]: [PASSED]
Feb 18, 2017 9:45:36 AM org.hawkular.cmdgw.ws.test.TestWebSocketClient$ExpectedEvent scheduleAnswer
FINE: No answer to send for message[1]
09:45:37,817 INFO  [org.hawkular.inventory.rest] (default task-38) HAWKINV002008: Sync with root [/t;itest-rest-tenant/f;d5091b6f-e3ee-4a30-a122-518ebf73bfaf/r;Local~~] finished in 16494ms
09:45:37,819 ERROR [org.hawkular.agent.monitor.protocol.EndpointService] (Hawkular WildFly Agent Full Discovery Scan-Local-1) HAWKMONITOR010049: Could not access resources of endpoint [DMREndpointService[Endpoint[Local]:[null]]]: java.lang.IllegalStateException: [org.hawkular.agent.monitor.protocol.dmr.DMREndpointService] must be in state [RUNNING] rather than [STOPPED] to perform [discoverChildren()]
	at org.hawkular.agent.monitor.service.ServiceStatus.assertRunning(ServiceStatus.java:34)
	at org.hawkular.agent.monitor.protocol.EndpointService.discoverChildren(EndpointService.java:339)
	at org.hawkular.agent.monitor.protocol.EndpointService.access$500(EndpointService.java:80)
	at org.hawkular.agent.monitor.protocol.EndpointService$1.run(EndpointService.java:303)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at org.jboss.threads.JBossThread.run(JBossThread.java:320)
```

The StandaloneDeployApplication is in the middle of running and its going to try to deploy an app (which includes running a discovery scan automatically so the new app can be immediately inventoried). However, the agent is partially (if not fully) shutdown by now and hasn't restarted yet. Hence the "must be in state [RUNNING] rather than [STOPPED]" message - the endpoint service cannot run discovery because it has been shutdown. At this point, the test is going to fail.

```
09:45:37,835 INFO  [org.hawkular.inventory.rest] (default task-61) HAWKINV002007: Commencing inventory sync on root entity [/t;itest-rest-tenant/f;d5091b6f-e3ee-4a30-a122-518ebf73bfaf/r;platform~%2FOPERATING_SYSTEM%3Dd5091b6f-e3ee-4a30-a122-518ebf73bfaf_OperatingSystem].
09:45:37,842 INFO  [org.hawkular.inventory.rest] (default task-61) HAWKINV002008: Sync with root [/t;itest-rest-tenant/f;d5091b6f-e3ee-4a30-a122-518ebf73bfaf/r;platform~%2FOPERATING_SYSTEM%3Dd5091b6f-e3ee-4a30-a122-518ebf73bfaf_OperatingSystem] finished in 6ms
09:45:37,847 INFO  [org.hawkular.agent.monitor.scheduler.SchedulerService] (Hawkular WildFly Agent Operation Start Thread) HAWKMONITOR010013: Stopping scheduler
09:45:37,876 WARN  [org.hawkular.agent.monitor.extension.OperationSubsystemStart] (Hawkular WildFly Agent Operation Start Thread) Starting Hawkular Monitor Service now.
09:45:37,880 INFO  [org.hawkular.agent.monitor.service.MonitorService] (Hawkular WildFly Agent Operation Start Thread) HAWKMONITOR010079: Agent being asked to start when it is already starting up. Will shutdown and try the startup again.
```

Somehow, the agent has been asked to start while it was already in the middle of starting so its going to kill the original start thread and try to start over. Things are really in a state that are tests do not expect. In any event, notice the agent is finally beginning to start up now - well after the UpdateCollectionIntervals test has finished and well into the StandaloneDeployApplication test.

jmazzitelli:
I think one solution is to change `isMonitoredServiceStarted` - rather than return a boolean we should return a state value enum "STARTING", "STARTED", "STOPPING", "STOPPED".

jmazzitelli:
> I think one solution is to change isMonitoredServiceStarted - rather than return a boolean we should return a state value enum "STARTING", "STARTED", "STOPPING", "STOPPED".

I'm in the process of coding something like this up. Hopefully, it might even mean I can get rid of all the startup threading we have going on.

jmazzitelli:
This should clean some stuff up during start and stop - this passes all tests for me locally - I think this is ready to be merged into this PR: https://github.com/jshaughn/hawkular-agent/pull/7

jmazzitelli:
Ok, I'm gonna do it. I'm gonna press the green button on this! I think everything is OK now.

Once I merge, it would be really good if @jshaughn you can build master branch and run some tests on your box to make sure your stuff still works and @josejulio if you can run a master branch build with EAP6 and make sure that works. This is all just to confirm that master branch is good.

I plan on running a full itest run on my box once I merge.

PR #291, opened 2017-02-11
WIP Prevent multiple restarts when changing multiple attributes on a …

josejulio:
…composite operation.

### DIFF
--- a/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java
+++ b/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java
@@ -17,20 +17,26 @@
 package org.hawkular.agent.monitor.extension;
 
 import java.util.Collection;
+import java.util.NoSuchElementException;
 
 import org.hawkular.agent.monitor.service.MonitorService;
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.OperationContext;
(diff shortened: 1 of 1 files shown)


josejulio:
Don't merge yet (although it might be pretty obvious that still need some refactoring)
This is something quickly that i though could prevent multiple restarts.
It does indeed prevent it. Tested on EAP7.1, single operations and composite operations modifying multiple attributes only trigger one restart (this time for sure).
WDYT @jmazzitelli ? 
//cc @jshaughn

jmazzitelli:
@josejulio  We should get some itests built so it always gets exercised and tested. @jshaughn already has tests that he's working on that exposed this problem, perhaps we just piggy back on what he's done for testing.

josejulio:
I have slightly modified it. (Removed a try-catch).

The problem with this approach is that I need to copy all the code from [applyUpdateToRuntime](https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/RestartParentWriteAttributeHandler.java#L54-L79), [revertUpdateFromRuntime](https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/RestartParentWriteAttributeHandler.java#L157-L172) (still need to do that) and other [private methods](https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/RestartParentWriteAttributeHandler.java#L174-L190) from the parent class.

I have thought on a second approach that should be cleaner and will send another PR with it. Following will be to add itests to check if the server is restarting more than once per operation (composite or not).

jmazzitelli:
PR #292 supercedes this PR... closing.

PR #292, opened 2017-02-11
Prevent multiple restarts when changing multiple attributes on a

josejulio:
composite operation.

I like this approach because I don't need to copy code from parent class.

@jmazzitelli please take a look at let me know if I should use this PR or the other as [base](https://github.com/hawkular/hawkular-agent/pull/291).

I'll continue on Monday, will add itests to check if the agent is being restarted more than once.

### DIFF
--- a/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java
+++ b/hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java
@@ -22,6 +22,7 @@ import org.hawkular.agent.monitor.service.MonitorService;
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
+import org.jboss.as.controller.OperationStepHandler;
 import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.RestartParentWriteAttributeHandler;
 import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
@@ -31,6 +32,12 @@ import org.jboss.msc.service.ServiceName;
 import org.jboss.msc.service.ServiceRegistry;
 
 public class MonitorServiceRestartParentAttributeHandler extends RestartParentWriteAttributeHandler {
+
+    private static final OperationContext.AttachmentKey<Integer> RECREATE_COUNTER =
+            OperationContext.AttachmentKey.create(Integer.class);
+
+    private static final OperationContext.AttachmentKey<Integer> REMOVE_COUNTER =
+            OperationContext.AttachmentKey.create(Integer.class);
 
     public MonitorServiceRestartParentAttributeHandler(AttributeDefinition... definitions) {
         super("subsystem", definitions);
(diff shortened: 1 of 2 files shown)


hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java
```diff
@@ -58,15 +65,39 @@ protected boolean requiresRuntime(OperationContext context) {
     @Override
     protected void removeServices(OperationContext context, ServiceName parentService, ModelNode parentModel)
```
josejulio, line 67: Let me know if you would like to have this on Util (or somewhere else) to execute blocks of code only once in a `OperationContext`.

jmazzitelli: I think this is good as is (assuming it all works :)) No need to put this in Util yet. Don't think we'll have a need for it - but if we do, we can refactor then.

I assume this works on EAP 6.4 as well as 7.0 and 7.1?

Let's see if we can get an itest in for this before merging. 

@pilhuhn  asked for an agent release today so we'll need to get something merged today to fix this problem before we release.

jshaughn:
Testing this locally the changes seems to work, so good job so far.  But the bad news is that because the restart is automatic the agent Command (in my test, UpdateCollectionIntervalsCommand) can not get its response sent back, because the agent is shutting down.

We need to have a way to get a Response to the client or the client won't know the command succeeded.

hawkular-wildfly-agent/src/main/java/org/hawkular/agent/monitor/extension/MonitorServiceRestartParentAttributeHandler.java
josejulio: EAP6.4 seems to be upset with something at this point. I'm trying to reason with it.

josejulio:
Umh, maybe the shutdown should wait for all the current jobs to finish or something like that?

jmazzitelli:
The problem is the restart is going to be completely separate from the command gateway command processing. We are getting commands in from the WebSocket, performing these DMR operations (which trigger this restart), but before the WebSocket processing receives the DMR operation response so it can send back the WebSocket response, the agent is restarted (which kills the WebSocket!).

Essentially, its a suicide mission - and the WebSocket processing kills itself before it knows it can send a response.

We need somehow to relay information from the DMR operation to the WebSocket command processor.

jmazzitelli:
@josejulio this is not to say you shouldn't get this PR working (with EAP 6 and 7) and eventually get it merged. The problem will need to be for us to workaround the WebSocket issue. It may mean enhancing the code in this PR, but this code in some form is still going to be needed.

pilhuhn:
On 13 Feb 2017, at 16:53, John Mazzitelli wrote:

> We need somehow to relay information from the DMR operation to the 
> WebSocket command processor.

Is there a way to dry-run the (composite) operation and
then send a response back upon the result?

That may be either failure = would fail for sure
or Looks good = it will most likely succeed.

And then we cache the information that we are running
the operation, so that after the restart we can send
a success message when we see the cached info

josejulio:
@jmazzitelli I got it to run on EAP6.4. 
Before (on EAP6.4) we needed to mark an step as Completed, thus. Will test on eap7.1 and push the code.

jshaughn:
@josejulio @jmazzitelli , I think I have a solution, it requires one change to this PR.  Currently We have this in `MonitorServiceRestartParentAttributeHandler:`

```
@Override
protected boolean isResourceServiceRestartAllowed(OperationContext context, ServiceController<?> service) {
    return true;
}
```

I think we want to remove this and go back to the default impl in the superclass.  What that means is that we don't actually perform a restart by default even if the operations perform changes that require a restart to be picked up.  It puts the responsibility on the caller to supply a header saying it's allowed.  For me this is a good thing, I don't really think we should always immediately restart, it could be the case that  the restart should be deferred.

To have you operation perform the automated restart you supply the operation header `ModelDescriptionConstants.ALLOW_RESOURCE_SERVICE_RESTART=true`.  This can be applied manually or with `OperationBuilder.allowResourceServiceRestart()`.

So that lets you perform an operation and control whether the restart kicks in.   This still leaves one problem, like we have in the new UpdateCollectionIntervals command.  Here we want a restart but because we are restarting the agent itself, the restart has to wait for the command response to be returned before getting invoked.  An immediate restart will prevent the response from being sent.  So, in this case we don't supply the `ALLOW_RESOURCE_SERVICE_RESTART` and instead invoke the restart manually from the command.  But we need it to be deferred.  This requires a small change in the agent's start command impl, which I have in a separate PR, but basically it puts the restart logic into a thread and offers a new "delay" attribute for the operation.

If you guys are OK with this I think we can move forward.  Either merge this as-is and I'll remove the  `isResourceServiceRestartAllowed()` override in my command PR, or update this first.  Otherwise, let me know what I missed...

jmazzitelli:
I'm going to merge this as-is. So we will have a baseline for this working "restart service" code.

We'll then work on the second PR that is needed for the update interval stuff.

PR #295, opened 2017-02-15
Reflect immutable + in-container attributes at server level.

pilhuhn:
While both are available at agent level, it is not what users expect and also make it harder for clients to quickly check if the server runs in container and/or is immutable.

This change also adds those 2 properties in the list of properties that the server sends to inventory at the level of the server resource.

bildschirmfoto 2017-02-15 um 10 43 23

### DIFF
--- a/eap6-support/hawkular-wildfly-agent-feature-pack-eap6/src/main/resources/subsystem-templates/hawkular-wildfly-agent.xml
+++ b/eap6-support/hawkular-wildfly-agent-feature-pack-eap6/src/main/resources/subsystem-templates/hawkular-wildfly-agent.xml
@@ -616,7 +616,7 @@
         <resource-config-dmr name="Bound Address"
                              path="/socket-binding-group=standard-sockets/socket-binding=http"
                              attribute="bound-address" />
-        <resource-config-dmr name="Is Immutable"
+        <resource-config-dmr name="Immutable"
                              path="/subsystem=hawkular-wildfly-agent"
                              attribute="immutable"
                              resolve-expressions="true"/>
--- a/eap6-support/hawkular-wildfly-agent-feature-pack-eap6/src/main/resources/subsystem-templates/hawkular-wildfly-agent.xml
+++ b/eap6-support/hawkular-wildfly-agent-feature-pack-eap6/src/main/resources/subsystem-templates/hawkular-wildfly-agent.xml
@@ -616,6 +616,14 @@
         <resource-config-dmr name="Bound Address"
                              path="/socket-binding-group=standard-sockets/socket-binding=http"
                              attribute="bound-address" />
+        <resource-config-dmr name="Is Immutable"
+                             path="/subsystem=hawkular-wildfly-agent"
+                             attribute="immutable"
+                             resolve-expressions="true"/>
+        <resource-config-dmr name="In Container"
+                             path="/subsystem=hawkular-wildfly-agent"
+                             attribute="in-container"
+                             resolve-expressions="true"/>
 
         <operation-dmr name="JDR"      internal-name="generate-jdr-report" path="/subsystem=jdr" />
         <operation-dmr name="Reload"   internal-name="reload" modifies="true">
--- a/hawkular-swarm-agents/hawkular-swarm-agent-dists/hawkular-swarm-agent-dist/src/main/resources/hawkular-swarm-agent-config.xml
+++ b/hawkular-swarm-agents/hawkular-swarm-agent-dists/hawkular-swarm-agent-dist/src/main/resources/hawkular-swarm-agent-config.xml
@@ -716,7 +716,7 @@
           <resource-config-dmr name="Bound Address"
                                path="/socket-binding-group=standard-sockets/socket-binding=http"
                                attribute="bound-address"/>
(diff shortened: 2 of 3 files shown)


jmazzitelli:
missing the same change to the EAP6 feature pack @pilhuhn  there is now 3 .xml files we have to keep in sync - the two you changed here, and the new EAP6 feature pack.

Also, consider adding these to the domain entities - (Domain server, host, controller???)

pilhuhn:
> Also, consider adding these to the domain entities - (Domain server, host, controller???)

DC/HC I get. DO you think the others also need this? @Jiri-Kremser You added domain support to MiQ - wdyt?

jmazzitelli:
> DC/HC I get. DO you think the others also need this?

We probably only need it for those top-level entities (so that's the WildFly Server for standalone (you already did those) and the Host Controller for domain mode).

But I'll let @Jiri-Kremser chime in with what he thinks since he knows better from the client UI point of view what is best.

jmazzitelli:
For reference, here's the EAP6 feature pack .xml that needs to change

https://github.com/hawkular/hawkular-agent/blob/master/eap6-support/hawkular-wildfly-agent-feature-pack-eap6/src/main/resources/subsystem-templates/hawkular-wildfly-agent.xml#L597

jkremser:
> DC/HC I get. DO you think the others also need this? @Jiri-Kremser You added domain support to MiQ - wdyt?

It's ok to have it only on the DC/HC. Having it also on the domain servers would be more convenient, but DC/HC is just fine.

pilhuhn:
@jmazzitelli Does the HC on EAP6 actually allow this? I think we are good on 7 though

jmazzitelli:
@pilhuhn the agent does NOT run in domain mode on EAP6. There is no way to run the agent inside the host controller on EAP6 - it did not allow for that. Being able to run your own subsystem extension in host controllers was added in later versions.

The agent (running external in another wildfly instance) can remotely monitor EAP6 domain mode, it just can't run inside of it in domain mode.

jmazzitelli:
@pilhuhn you'll still want to add this immutable and in container flag in the standalone WildFly Server resource type in the EAP6 feature pack .xml.

jmazzitelli:
I take that back.. does NOT LGTM :)

@pilhuhn  - you should change the agent property name to be consistent - the agent's prop is called "Immutable"... these ones you added are called "Is Immutable"

You should make them all the same. I don't care which, but just make them all the same

eap6-support/hawkular-wildfly-agent-feature-pack-eap6/src/main/resources/subsystem-templates/hawkular-wildfly-agent.xml
```diff
@@ -616,6 +616,14 @@
         <resource-config-dmr name="Bound Address"
                              path="/socket-binding-group=standard-sockets/socket-binding=http"
                              attribute="bound-address" />
+        <resource-config-dmr name="Immutable"
+                             path="/subsystem=hawkular-wildfly-agent"
```
josejulio, line 620: @jmazzitelli What will happen when we are monitoring a remote server (via remote-dmr) ?
Those won't always have hawkular-wildfly-agent.

jmazzitelli: good point. these will be null - I can't remember if the agent will still create this property or not. But you will most likely get a warning or error in the log file.

jmazzitelli: @pilhuhn - are you ok with that behavior? If the WildFly is being remotely monitored (that is, the agent is NOT deployed inside of the wildfly it is monitoring) these values will not be set or be set to null (one of those two, can't remember which).

pilhuhn:
On 15 Feb 2017, at 18:20, John Mazzitelli wrote:

> jmazzitelli commented on this pull request.
>
>
>
>> @@ -616,6 +616,14 @@
>          <resource-config-dmr name="Bound Address"
>                               path="/socket-binding-group=standard-sockets/socket-binding=http"
>                               attribute="bound-address" />
> +        <resource-config-dmr name="Immutable"
> +                             path="/subsystem=hawkular-wildfly-agent"
>
> @pilhuhn - are you ok with that behavior? If the WildFly is being 
> remotely monitored (that is, the agent is NOT deployed inside of the 
> wildfly it is monitoring) these values will not be set or be set to 
> null (one of those two, can't remember which).

Probably something we need to then catch in client
code and react accordingly.
I would even think that in that case we should treat the
remote as immutable as well.
Also the original property on the agent will have the
same issue - no?

jmazzitelli:
Because its the agent itself, the property will always have a value (since the agent is always there). This means "this agent is immutable and you cannot change its configurations, nor can you operate on any of its monitored servers (which may be remote)."

But if the agent is immutable (i.e. deployed in a container) I think it is a fair assuming that all its local and remote managed servers are also immutable (since they are most likely also deployed in a container/OpenShift/whatever).

PR #298, opened 2017-02-20
This test fails spuriously. I think due to test timing. This might fix it.

jmazzitelli:


jshaughn:
Fwiw, I have seen that test sink travis and also fail occasionally locally.  Hopefully this fixes it.

PR #299, opened 2017-02-21
[HWKAGENT-161] Add option to installer to make config changes without…

pgier:
… module install

### DIFF
--- a/hawkular-wildfly-agent-itest-parent/hawkular-wildfly-agent-all-itests/hawkular-wildfly-agent-installer-itest/pom.xml
+++ b/hawkular-wildfly-agent-itest-parent/hawkular-wildfly-agent-all-itests/hawkular-wildfly-agent-installer-itest/pom.xml
@@ -147,6 +147,17 @@
                 </configuration>
               </execution>
               <execution>
+                <id>server-provisioning-wildfly-hawkular-agent-module</id>
+                <goals>
+                  <goal>build</goal>
+                </goals>
+                <phase>compile</phase>
+                <configuration>
+                  <config-file>server-provisioning-plain-wildfly.xml</config-file>
+                  <server-name>wildfly-hawkular-agent-module</server-name>
+                </configuration>
+              </execution>
+              <execution>
                 <id>server-provisioning-hawkular</id>
                 <goals>
                   <goal>build</goal>
@@ -155,6 +166,32 @@
                 <configuration>
                   <config-file>server-provisioning-hawkular.xml</config-file>
                   <server-name>hawkular</server-name>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-dependency-plugin</artifactId>
+            <executions>
+              <execution>
+                <id>extract-hawkular-agent-module</id>
+                <goals>
+                  <goal>unpack</goal>
+                </goals>
+                <phase>compile</phase>
+                <configuration>
+                  <artifactItems>
+                    <artifactItem>
+                      <groupId>org.hawkular.agent</groupId>
+                      <artifactId>hawkular-wildfly-agent-wf-extension</artifactId>
+                      <version>${project.version}</version>
+                      <type>zip</type>
+                      <overWrite>true</overWrite>
+                      <outputDirectory>${project.build.directory}/wildfly-hawkular-agent-module/modules</outputDirectory>
+                      <destFileName>optional-new-name.jar</destFileName>
+                    </artifactItem>
+                  </artifactItems>
                 </configuration>
               </execution>
             </executions>
@@ -214,6 +251,8 @@
                 <plain-wildfly.bind.address>${plain-wildfly.bind.address}</plain-wildfly.bind.address>
                 <plain-wildfly.management.http.port>${plain-wildfly.management.http.port}</plain-wildfly.management.http.port>
                 <plain-wildfly.port.offset>${plain-wildfly.port.offset}</plain-wildfly.port.offset>
+
+                <wildfly-hawkular-agent-module.home.dir>${project.build.directory}/wildfly-hawkular-agent-module</wildfly-hawkular-agent-module.home.dir>
 
                 <!-- output over-the-wire traffic -->
                 <org.apache.commons.logging.Log>
--- a/hawkular-wildfly-agent-itest-parent/hawkular-wildfly-agent-all-itests/hawkular-wildfly-agent-installer-itest/src/test/java/org/hawkular/wildfly/agent/installer/AgentInstallerStandalonePrepare.java
+++ b/hawkular-wildfly-agent-itest-parent/hawkular-wildfly-agent-all-itests/hawkular-wildfly-agent-installer-itest/src/test/java/org/hawkular/wildfly/agent/installer/AgentInstallerStandalonePrepare.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
+ * Copyright 2015-2017 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -52,7 +52,7 @@ public class AgentInstallerStandalonePrepare extends AbstractITest {
     @Test
     public void installAgent() throws Throwable {
 
-        assertAgentInStandalone(false);
+        assertAgentInStandalone(false, wfHome);
         File agentModuleXml = new File(wfHome,
                 "modules/system/add-ons/hawkular-agent/org/hawkular/agent/main/module.xml");
         Assert.assertFalse(agentModuleXml.exists(), "[" + agentModuleXml.getAbsolutePath() + "] should not exist");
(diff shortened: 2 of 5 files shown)


pgier:
Not sure if this is the best way to implement this, just submitting for discussion.
And this PR does not currently include tests.

jmazzitelli:
This looks fine to me the way it is implemented. We have several installer tests in itests and since those all pass, at least nothing was broken.

One way you could test this is add something to https://github.com/hawkular/hawkular-agent/blob/master/hawkular-wildfly-agent-itest-parent/hawkular-wildfly-agent-all-itests/hawkular-wildfly-agent-installer-itest/src/test/java/org/hawkular/wildfly/agent/installer/AgentInstallerStandalonePrepare.java (or create a new test class that does something similar) - your new test would ensure it runs after this one (this AgentInstallerStandalonePrepare installs the agent - both config and add-ons). So you would install after that code runs.

The problem is, installing on top of WildFly that already had an agent &lt;subsystem> would cause problems (I've seen it where it adds a second &lt;subsystem>). Perhaps this is the time to get the agent installer to fix that issue and let it install config over one where the agent config already exists.

hawkular-wildfly-agent-itest-parent/hawkular-wildfly-agent-all-itests/hawkular-wildfly-agent-installer-itest/pom.xml
```diff
@@ -147,6 +147,17 @@
...
               </execution>
               <execution>
+                <id>server-provisioning-wildfly-hawkular-agent-module</id>
+                <goals>
+                  <goal>build</goal>
+                </goals>
+                <phase>compile</phase>
+                <configuration>
+                  <config-file>server-provisioning-plain-wildfly.xml</config-file>
+                  <server-name>wildfly-hawkular-agent-module</server-name>
```
jmazzitelli, line 157: Give this server-name something more descriptive to explain what it is for. I had a tough time understanding until I examined things closely. But if you called this "wildfly-for-config-only-install-tests" it would have been immediately clear.

hawkular-wildfly-agent-itest-parent/hawkular-wildfly-agent-all-itests/hawkular-wildfly-agent-installer-itest/pom.xml
```diff
@@ -159,6 +170,32 @@
               </execution>
             </executions>
           </plugin>
+          <plugin>
```
jmazzitelli, line 173: Can you put a comment here to explain what this is doing? Also, does the comment in line 133 need to amended now?

hawkular-wildfly-agent-itest-parent/hawkular-wildfly-agent-all-itests/hawkular-wildfly-agent-installer-itest/pom.xml
```diff
@@ -159,6 +170,32 @@
...
+                <configuration>
+                  <artifactItems>
+                    <artifactItem>
+                      <groupId>org.hawkular.agent</groupId>
+                      <artifactId>hawkular-wildfly-agent-wf-extension</artifactId>
+                      <version>${project.version}</version>
+                      <type>zip</type>
+                      <overWrite>true</overWrite>
+                      <outputDirectory>${project.build.directory}/wildfly-hawkular-agent-module/modules</outputDirectory>
+                      <destFileName>optional-new-name.jar</destFileName>
```
jmazzitelli, line 192: What is this "optional-new-name.jar" destination file? Is that correct? What is it for? If this is correct, a comment would be beneficial for a reader to understand why this is here/what it is for.

hawkular-wildfly-agent-itest-parent/hawkular-wildfly-agent-all-itests/hawkular-wildfly-agent-installer-itest/pom.xml
```diff
@@ -159,6 +170,32 @@
...
+                <phase>compile</phase>
+                <configuration>
+                  <artifactItems>
+                    <artifactItem>
+                      <groupId>org.hawkular.agent</groupId>
+                      <artifactId>hawkular-wildfly-agent-wf-extension</artifactId>
+                      <version>${project.version}</version>
+                      <type>zip</type>
+                      <overWrite>true</overWrite>
+                      <outputDirectory>${project.build.directory}/wildfly-hawkular-agent-module/modules</outputDirectory>
```
jmazzitelli, line 191: for the outputDirectory, use the new property introduced on line 255 (which should have its "brother" property defined somewhere around line 118 - see how the rest of the properties are defined there and are propagated down to the systemPropertyVariables configuration). Note also I recommend a change in the name to be more descriptive:

```
<outputDirectory>${wildfly-for-config-only-install-tests.home.dir}/modules</outputDirectory>
```

hawkular-wildfly-agent-itest-parent/hawkular-wildfly-agent-all-itests/hawkular-wildfly-agent-installer-itest/pom.xml
```diff
@@ -147,6 +147,17 @@
                 </configuration>
               </execution>
               <execution>
+                <id>server-provisioning-wildfly-hawkular-agent-module</id>
```
jmazzitelli, line 150: Give this id something more descriptive to explain what it is for. I had a tough time understanding until I examined things closely. But if you called this "server-provisioning-wildfly-for-config-only-install-tests" it would have been immediately clear.

hawkular-wildfly-agent-itest-parent/hawkular-wildfly-agent-all-itests/hawkular-wildfly-agent-installer-itest/pom.xml
```diff
@@ -215,6 +252,8 @@
                 <plain-wildfly.management.http.port>${plain-wildfly.management.http.port}</plain-wildfly.management.http.port>
                 <plain-wildfly.port.offset>${plain-wildfly.port.offset}</plain-wildfly.port.offset>
 
+                <wildfly-hawkular-agent-module.home.dir>${project.build.directory}/wildfly-hawkular-agent-module</wildfly-hawkular-agent-module.home.dir>
```
jmazzitelli, line 255: Rename this to be more descriptive ("wildfly-for-config-only-install-tests.home.dir"). Also, define this like the rest (i.e. put a property of the same name around line 118):

```
<wildfly-hawkular-agent-module.home.dir>${wildfly-hawkular-agent-module.home.dir}</wildfly-hawkular-agent-module.home.dir>
```

This will let you reuse it up on line 191 and thus won't have these paths hardcoded in multiple places (i.e. here and line 191).

hawkular-wildfly-agent-itest-parent/hawkular-wildfly-agent-all-itests/hawkular-wildfly-agent-installer-itest/src/test/java/org/hawkular/wildfly/agent/installer/AgentInstallerStandalonePrepare.java
```diff
@@ -72,11 +72,38 @@ public void installAgent() throws Throwable {
...
 
-    private void assertAgentInStandalone(boolean agentAvailableExpected) throws UnsupportedEncodingException,
+    @Test
+    public void installAgentConfigOnly() throws Throwable {
+
+        String agentExtensionZipPath = System.getProperty("hawkular-wildfly-agent-wf-extension.zip.path");
+        Assert.assertTrue(new File(agentExtensionZipPath).exists(),
+                "${hawkular-wildfly-agent-wf-extension.zip.path} [" + agentExtensionZipPath + "] does not exist");
+
+        String wfHomeProperty = System.getProperty("wildfly-hawkular-agent-module.home.dir");
```
jmazzitelli, line 86: The name of this sys prop (i.e. the argument passed to System.getProperty) will need to be changed - match the new descriptive name (see my github comments on the pom.xml)

```
String wfHomeProperty = System.getProperty("wildfly-for-config-only-install-tests.home.dir");
```

wildfly-module-installer/src/main/java/org/hawkular/wildfly/module/installer/ExtensionDeployer.java
```diff
@@ -32,6 +34,10 @@
     private File sourceServerConfigBackupAbsolute;
     private File modulesHomeAbsolute;
 
+    private static final String DEFAULT_MODULE_ID = "org.hawkular.agent";
```
jmazzitelli, line 37: Here is the problem with this hardcoding of these constants. This entire "wildfly-module-installer" artifact (notice it doesn't have the name "hawkular" in it) which this code belongs is not actually hawkular wildfly agent specific. This was originally written to support deploying anything in WildFly (not even hawkular related code) and may in fact be being used for non-hawkular agent install things. I do not know if anyone else is using this artifact for things other than hawkular wildfly agent - but we should not now pin this down with hardcoded strings related to the agent which would prohibit us or others from using this code in the future for non-agent things.

You should parameterize these two things (default module ID and default module relative path) such that they are not hard-coded in here but rather in the hawkular wildfly agent installer artifact code. Put them in the DeploymentConfiguration like the other things so AgentInstaller can pass these values to the DeploymentConfiguration builder (see AgentInstaller line 122 or thereabouts). I envision being able to do something like this:

```
             Builder configurationBldr = DeploymentConfiguration.builder()
                     .jbossHome(new File(jbossHome))
                     .module(moduleZipUrl)
                     .defaultModuleId("org.hawkular.agent") // <--- NEW
                     .defaultModuleRelativePath("/system/add-ons/hawkular-agent") // <---NEW
                     .socketBinding(socketBindingSnippetFile.toURI().toURL());

 ```

jmazzitelli (changes requested):

jmazzitelli:
@pgier most of my requested changes are minor and should be trivial. The only substantial change I asked for is the change to the DeployerConfiguration so as to keep the wildfly-module-installer artifact code hawkular-agnostic. But that too should be fairly easy to change.

Other than that, this code should be good and can be merged when those changes are made. Travis shows "red" but I looked at the raw log and the tests did fully pass - its just the underlying cassandra server/h-metrics failed to shutdown timely and that causes travis to think it failed. Once you submit the changes, we'll see if travis goes green which I suspect it will.

jmazzitelli:
I was told that @pgier is away so rather than wait, I made the requested changes on top of this PR in my own PR #301. If all goes green, I'll merge that PR and close this one. If I run into problems, I'll wait to hear back from @pgier.

jmazzitelli:
Since Paul is out for a bit, rather than submit a PR on his PR, I simply took this PR and put my changes on top and squash merged it. See PR #301. That passed all tests, travis was green. So this PR can be closed since it has been superceded by PR #301.

PR #302, opened 2017-03-06
Refactor Agent Engine Core

jmazzitelli:
This refactors the Hawkular WildFly Agent core engine such that it is no longer a requirement that the agent be deployed within an EAP or WildFly server nor is it required to be deployed as a WildFly subsystem extension. The agent can still be deployed in that manner (as a WildFly subsystem extension) however the core engine can now be embedded in any Java VM. More importantly, it can now be injected in any VM via the Java -javaagent comment line option.

All the features still exist no matter the mode of deployment - specifically, the agent can still collect DMR and JMX inventory and metrics (or just metrics if you put the agent in metrics only mode).

The engine (org.hawkular.agent.javaagent.JavaAgentEngine) can be instantiated by passing to its constructor an agent engine configuration file (which is in YAML format).

See the README for instructions on injecting as a javaagent.


jmazzitelli:
Created a JIRA for this: https://issues.jboss.org/browse/HWKAGENT-162

PR #304, opened 2017-03-18
HWKAGENT-163 support ${x} tokens in the settings that need to be configurable

jmazzitelli:


jmazzitelli:
YAML now supports ${x} where "x" is a system property. You can optionally define default values as in ${x:default}. If you prefix "x" with "ENV~" then "x" is considered an environment variable and its value will be looked up in the process environment. For example ${ENV\~HOME}. You can still use defaults when using environment variables, for example: ${ENV\~SOME_VAR:default}.

Note that only some YAML properties support expressions - the ones typically you would want to configure with sys props or env vars.

PR #306, opened 2017-03-27
HAWKULAR-1207 container id property

jmazzitelli:


jshaughn:
LGTM as long as containerIdFile is guaranteed to be "/proc/self/cgroup".

abonas:
@rubenvp8510 can you please review

rubenvp8510:
LGTM

jmazzitelli:
After peer review and some of my own local testing, this looks to be OK to merge.

There is now a container-id (that exists in the same manner, and is set in the same manner) as machine-id - in other words, there is now a "container-id" property on the "platform" configuration element in the agent configuration file (analogous to the machine-id property); you can set it there or you can set it via the system property hawkular.agent.container.id (analogous to hawkular.agent.machine.id). If you do not set it, the agent will look for a docker container ID in /proc/self/cgroup and if it finds one there it will use it.

PR #307, opened 2017-03-29
try harder to figure out what the local DMR address/port is

jmazzitelli:
This just attempts to make a better guess at the local-dmr address and port of the management endpoint.

Since we are running in the same VM we should be able to get the system properties that determine any custom address and port settings for the management endpoint of the local WildFly/EAP.


jmazzitelli:
@jshaughn review this - we can merge once someone else says it looks good.

The idea here is the Java Agent wants to guess what the local DMR endpoint is - if someone overrides the address, port, or port-offset via startup system properties, the agent should be able to detect those and use them rather than rely on the hardcoded defaults (which are 127.0.0.1, 9090, and 0 respectively).

jshaughn:
This should help guess better.  It won't detect a change they make by editing the config file itself, but I would hope most people would use the sysprop. LGTM.

jmazzitelli:
This is for HWKAGENT-164

PR #308, opened 2017-03-30
Creates hawkular-javaagent-wildfly-feature-pack

josejulio:
A Wildfly feature pack with the Hawkular java agent built-in.

Adding `hawkular.rest.host` ENV to make it easier to define a custom URL.

I'm going to use this one to create a hawkular-services based dist with HJA instead of the Wildfly Java agent.


hawkular-javaagent-wildfly-dist/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+              <config-file>server-provisioning-plain-wildfly.xml</config-file>
+              <server-name>${project.build.finalName}</server-name>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+
+      <!-- Install the Hawkular WildFly Agent inside the WildFly server -->
+      <!--
+           The agent will be installed in standalone.xml file as well as host.xml and domain.xml.
```
jmazzitelli: I assume this comment is not right. The agent doesn't go in standalone.xml, host.xml, or domain.xml.

hawkular-javaagent-wildfly-feature-pack/feature-pack-build.xml
```diff
@@ -0,0 +1,47 @@
...
+    limitations under the License.
+
+-->
+<build xmlns="urn:wildfly:feature-pack-build:1.1">
+  <dependencies>
+    <artifact name="org.wildfly:wildfly-feature-pack" />
+  </dependencies>
+
+  <config>
+    <standalone template="configuration/standalone/template.xml" subsystems="configuration/standalone/subsystems.xml" output-file="standalone/configuration/standalone.xml" />
```
jmazzitelli: why is standalone.xml getting changed here? The Java Agent should be able to be used even if standalone.xml is the out-of-box one that comes with WildFly with no changes needed.

hawkular-javaagent-wildfly-feature-pack/pom.xml
```diff
@@ -0,0 +1,144 @@
...
+            <goals>
+              <goal>unpack</goal>
+            </goals>
+            <configuration>
+              <artifactItems>
+                <artifactItem>
+                  <groupId>org.wildfly</groupId>
+                  <artifactId>wildfly-feature-pack</artifactId>
+                  <type>zip</type>
+                  <includes>configuration/**/*.xml</includes>
```
jmazzitelli: What .xml files are needed for this?

hawkular-javaagent-wildfly-feature-pack/feature-pack-build.xml
josejulio: It isn't changed, I'm just using the defaults template.xml and subsystems.xml to generate the standalone.xml

josejulio: I suppose that there is some other way to get them? I'll take a look at it.

hawkular-javaagent-wildfly-dist/pom.xml
josejulio: Yep, totally incorrect comment.

hawkular-javaagent-wildfly-feature-pack/feature-pack-build.xml
jmazzitelli: I am just wondering where `configuration/standalone/template.xml` and `configuration/standalone/subsystems.xml` are. I thought you had your own versions of those .xml files in here (but I don't see them). As long as we aren't maintaining our own here in this maven module. If you don't have your own copy, I assume they are coming from the wildfly feature pack?

hawkular-javaagent-wildfly-feature-pack/pom.xml
josejulio: Actually just these two files are needed to generate standalone.xml:
`configuration/standalone/template.xml` and `configuration/standalone/subsystems.xml`

hawkular-javaagent-wildfly-feature-pack/feature-pack-build.xml
josejulio: You assumption is correct, see https://github.com/hawkular/hawkular-agent/pull/308#discussion_r109006882

hawkular-javaagent-wildfly-feature-pack/pom.xml
jmazzitelli: OK, I see how it works now :)  This looks good.

hawkular-javaagent-wildfly-dist/pom.xml
josejulio: Totally incorrect comment is gone, plus updated the pom description.

hawkular-javaagent-wildfly-feature-pack/pom.xml
josejulio: After reading a bit more, I don't think i need this part. Will test if needed and possibly remove it.

josejulio: Well, i was wrong. I still need those files.
Currently I'm only including the ones that are required.

josejulio:
So I had this standalone.conf on the code. 
Thanks to @ppalaga I managed to remove it from the code.

hawkular-javaagent-wildfly-feature-pack/pom.xml
```diff
@@ -68,12 +74,64 @@
...
+                      tofile="${project.build.directory}/feature-pack-resources/content/bin/standalone.conf"
+                      overwrite="true"/>
+                <replaceregexp file="${project.build.directory}/feature-pack-resources/content/bin/standalone.conf"
+                               match="(JBOSS_MODULES_SYSTEM_PKGS=&quot;.*)&quot;"
+                               replace="\1,org.jboss.logmanager&quot;"
+                               flags="s"
+                               byline="true"
+                />
+                <concat destfile="${project.build.directory}/feature-pack-resources/content/bin/standalone.conf" append="yes">
+                  <file name="${basedir}/src/main/bin/standalone-hawkular-javaagent-javaopts.conf"/>
```
jmazzitelli: shouldn't you put that file under the standard "src/main/resources" location? I've never see the convention of putting files in "src/main/bin".

josejulio: Resources are copied to the feature pack. But you are right, it looks rare to have those files over there. I'll move feature pack resources under resources/feature-pack and move that standalone to resources/standalone-haw...

pilhuhn:
On 30 Mar 2017, at 20:39, Josejulio Martínez wrote:
> It isn't changed, I'm just using the defaults template.xml and 
> subsystems.xml to generate the standalone.xml

Is that even needed? I understood the javaagent would
use an external .yml file for configuration.
We should probably not touch the standalone.xml
coming from the server at all.

jmazzitelli:
> We should probably not touch the standalone.xml
> coming from the server at all.

That is what I was getting at. I agree. We should not touch any file (other than standalone.conf) coming from the stock, out-of-box, WildFly.

josejulio:
> Is that even needed? I understood the javaagent would
> use an external .yml file for configuration.
> We should probably not touch the standalone.xml
> coming from the server at all.

I'm not really touching it, but as far as i could tell there isn't any `standalone.xml`. I need to specify how it will be created. I'm using the same rule used on the [wildfly-feature-pack](https://github.com/wildfly/wildfly/blob/master/feature-pack/feature-pack-build.xml#L29-L32).

josejulio:
> We should probably not touch the standalone.xml
> coming from the server at all.
> That is what I was getting at. I agree. We should not touch any file (other than standalone.conf) coming from the stock, out-of-box, WildFly.

Ok, I guess i found out what you meant by that and how to fix it. 
I'll remove that from the feature-pack and `wildfly-feature-pack` to the provisioning file.

hawkular-javaagent-wildfly-dist/server-provisioning-plain-wildfly.xml
```diff
@@ -0,0 +1,24 @@
...
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<server-provisioning xmlns="urn:wildfly:server-provisioning:1.1" extract-schemas="true" copy-module-artifacts="true">
+  <feature-packs>
+    <feature-pack groupId="org.hawkular.agent" artifactId="hawkular-javaagent-wildfly-feature-pack" version="${project.version}" />
+    <feature-pack groupId="org.wildfly" artifactId="wildfly-feature-pack" version="${version.org.wildfly}" />
```
josejulio, line 22: Use this feature-pack which has the standalone.xml file.
We use it after `hawkular-javaagent-wildfly-feature-pack` to keep `standalone.conf` from being overwritten.

hawkular-javaagent-wildfly-feature-pack/src/main/resources/featurepack/content/standalone/configuration/hawkular-javaagent-config.yaml
```diff
@@ -0,0 +1,1462 @@
...
+    - name: Status
+      internal-name: status
+    - name: Inventory Discovery Scan
+      internal-name: fullDiscoveryScan
+    - name: Inventory Report
+      internal-name: inventoryReport
+
+# MANAGED SERVERS
+
+managed-servers:
```
jmazzitelli, line 1404: @josejulio the agent now supports the ability to wait for certain resources before starting to monitor the server (see https://issues.jboss.org/browse/HWKAGENT-165). This helps with the problem of the delay=XX not being long enough and the agent starting to monitor too early. So for these local-dmr, remote-dmr entries, add a "wait-for" to wait for the "/" resource. For the local-jmx and remote-jmx, we can put in a "wait-for" on the mbean "java.lang:type=Runtime". It looks like:

```
local-dmr:
  ...the other config properties here...
  wait-for:
  - name: /
```

or for jmx:

```
local-jmx:
  ...the other config properties here...
  wait-for:
  - name: java.lang:type=Runtime
```

Same thing for the remote-dmr and remote-jmx.

josejulio: I'll update the config and test it. I suppose I can remove the delay or at least set to a small value?

jmazzitelli: Set it to a small value. We can use it as documentation. :) People will see it and know they can set it higher if they need to.

jmazzitelli:
@josejulio I'm not even sure if people are going to use this, but perhaps we should also create an EAP6 feature pack called `hawkular-javaagent-feature-pack-eap6` and put it under the "eap6-support" directory here: https://github.com/hawkular/hawkular-agent/tree/master/eap6-support

Should be simple - just copy what you got for your current feature pack but it would use the EAP6 yaml config file - https://github.com/hawkular/hawkular-agent/blob/master/hawkular-javaagent/src/test/resources/real-config-eap6.yaml

josejulio:
Sure, I actually though there weren't feature-packs for EAP6 builds.

josejulio:
@jmazzitelli I just updated the code.
Added the wait-for to the config and create the EAP6 feature-pack with the EAP6 config file (also with the wait-for).

jmazzitelli:
I did a manual squash merge of this - so this has been merged, I just did it manually and because I did a rebase with squash looks like github can't see it. So I will close this. But this is the commit where this PR went into master: https://github.com/hawkular/hawkular-agent/commit/6cc93d5f28bd84f56d08164f9b8cbfd534a4c0cb

The reason I did it manually was because the pom.xml files in this PR had a version of 0.29.3.Final-SNAPSHOT but we are at 0.29.4.Final-SNAPSHOT - so I just updated the versions to the latest, squash merged, and pushed.

PR #313, opened 2017-04-05
HWKAGENT-166 Adds Dockerfile for Wildfly with Hawkular-javaagent

josejulio:
~~I will migrate the other Dockerfiles to use rely on maven for downloading the dependencies (wildfly agent) and building with the do.sh~~

Fixed a couple of "bugs" on the domain Dockerfile that made fail the installation of the Hawkular Wildfly Agent.

Could you please take a look @pilhuhn @jmazzitelli ?


docker-dist/Dockerfile
```diff
@@ -19,29 +19,22 @@ FROM jboss/wildfly:10.1.0.Final
 
 MAINTAINER Heiko W. Rupp <hawkular-dev@lists.jboss.org>
```
pilhuhn: Can you change my name to Hawkular project please while in that file?

docker-dist/Dockerfile
```diff
@@ -19,29 +19,22 @@ FROM jboss/wildfly:10.1.0.Final
...
+ADD target/hawkular-javaagent-wildfly-dist/bin/standalone.conf $JBOSS_HOME/bin
+ADD target/hawkular-javaagent-wildfly-dist/bin/hawkular-javaagent.jar $JBOSS_HOME/bin
+ADD target/hawkular-javaagent-wildfly-dist/standalone/configuration/hawkular-javaagent-config.yaml $JBOSS_HOME/standalone/configuration/
 
+ENV HAWKULAR_SERVER_PROTOCOL http
 ENV HAWKULAR_SERVER_ADDR hawkular
 ENV HAWKULAR_SERVER_PORT 8080
 ENV HAWKULAR_AGENT_USER=jdoe HAWKULAR_AGENT_PASSWORD=password
-ENV JAVA_OPTS="-Xmx256m -Djava.net.preferIPv4Stack=true -Djboss.modules.system.pkgs=org.jboss.byteman -Djava.awt.headless=true -XX:MaxMetaspaceSize=256m"
+ENV JAVA_OPTS="-Xmx256m -Djava.net.preferIPv4Stack=true -Djboss.modules.system.pkgs=org.jboss.byteman,org.jboss.logmanager -Djava.awt.headless=true -XX:MaxMetaspaceSize=256m"
```
pilhuhn: The javaagent is missing here?

pilhuhn (changes requested):

docker-dist/Dockerfile
josejulio: Is included [here](https://github.com/hawkular/hawkular-agent/pull/313/files/149416054c57bceffd9b2fca21227bce7d62d421#diff-f235893104df7b8b55fa6ebf8e49c6d1R23) in the custom [standalone.conf](https://github.com/hawkular/hawkular-agent/blob/master/hawkular-javaagent-wildfly-feature-pack/src/main/resources/bin/standalone-hawkular-javaagent-javaopts.conf#L18-L21) from the feature-pack.

docker-dist/Dockerfile
josejulio: Yes, I'll modify the other Dockerfiles while I'm on it. Is that ok? Or just this file?

pilhuhn: Both is good.

docker-dist/Dockerfile
pilhuhn: Ok, thought so, but was confused because you put the logmanager here and not in standalone.conf.
I think it may be good to add this in standalone.conf with some check if it was set before, as the opts that are passed here come from env and the user could do `docker -e JAVA_OPTS=frobnitz` and thus lose the logmanager , which may end up with strange errors

josejulio: You are right on this, I was having that issue. I'll do something about it.

hawkular-javaagent-wildfly-feature-pack/pom.xml
```diff
@@ -95,7 +95,20 @@
...
+                               flags="s"
+                               byline="true" />
+                <!--   If JAVA_OPTS is set, we need to notify that JBOSS_MODULES_SYSTEM_PKGS is to be set if we want to modify it.
+                       Include $JBOSS_MODULES_SYSTEM_PKGS in $JAVA_OPTS-->
+                <replaceregexp file="${project.build.directory}/feature-pack-resources/content/bin/standalone.conf"
+                               match="( *)echo &quot;JAVA_OPTS already set in environment; overriding default settings with values: \$JAVA_OPTS&quot;"
+                               replace="\1echo &quot;JAVA_OPTS already set in environment; overriding default settings with values: $JAVA_OPTS&quot;&#10;\1echo &quot;If you want to set 'jboss.modules.system.pkgs' use JBOSS_MODULES_SYSTEM_PKGS env&quot;&#10;\1JAVA_OPTS=&quot;$JAVA_OPTS -Djboss.modules.system.pkgs=$JBOSS_MODULES_SYSTEM_PKGS&quot;"/>
```
josejulio: This block is hard to read, I'm thinking on using groovyscript to replace the `replaceregexp`.
WDYT? Should i leave it like this or do something about it? 

The `replaceregexp` is used to generate the following section of the standalone.conf.

```bash
if [ "x$JBOSS_MODULES_SYSTEM_PKGS" = "x" ]; then
   JBOSS_MODULES_SYSTEM_PKGS="org.jboss.byteman,org.jboss.logmanager"
   else
      echo $JBOSS_MODULES_SYSTEM_PKGS | grep "org.jboss.logmanager" > /dev/null
      if ! [ $? -eq 0 ]; then
         JBOSS_MODULES_SYSTEM_PKGS="$JBOSS_MODULES_SYSTEM_PKGS,org.jboss.logmanager"
      fi
fi

# ...

if [ "x$JAVA_OPTS" = "x" ]; then
   JAVA_OPTS="-Xms64m -Xmx512m -XX:MetaspaceSize=96M -XX:MaxMetaspaceSize=256m -Djava.net.preferIPv4Stack=true"
   JAVA_OPTS="$JAVA_OPTS -Djboss.modules.system.pkgs=$JBOSS_MODULES_SYSTEM_PKGS -Djava.awt.headless=true"
else
   echo "JAVA_OPTS already set in environment; overriding default settings with values: $JAVA_OPTS"
   echo "If you want to set 'jboss.modules.system.pkgs' use JBOSS_MODULES_SYSTEM_PKGS env"
   JAVA_OPTS="$JAVA_OPTS -Djboss.modules.system.pkgs=$JBOSS_MODULES_SYSTEM_PKGS"
fi
```

pilhuhn: I don't think using Groovy is a good idea as it adds yet another dependency into the container. Or do you mean inside the pom?
Why not just write a hand-crafted standalone.conf and copy that file literally without all that maven/ant magic?

josejulio: I meant on compile-time (inside the pom). But if you don't see anything wrong with that `replaceregexp` then is probably fine as is.

Updated the first comment, i meant that the `replaceexp` might be hard to read.

josejulio: > Why not just write a hand-crafted standalone.conf and copy that file literally without all that maven/ant magic?

Not sure if anything relevant might come from there. IIRC @ppalaga suggested that  a hand-crafted `standalone.conf` might be harder to maintain, but to be honest, not sure if those regexps are anything easier :).
If you are fine with that I could use a hand-crafted `standalone.conf`.

josejulio:
I'm now using a fixed standalone.conf to avoid all ant/maven sorcery.

pilhuhn (approved):

pilhuhn:
@jmazzitelli Are you happy too?

jmazzitelli:
@josejulio @pilhuhn did I miss it? Where do we set the immutable flag and in-container flag to "true"?

josejulio:
I though i set them somewhere... but can't find it.

I set it up using 
```Dockerfile
CMD /opt/jboss/wildfly/bin/standalone.sh -b 0.0.0.0 \
-Dhawkular.rest.host=${HAWKULAR_SERVER_PROTOCOL}://${HAWKULAR_SERVER_ADDR}:${HAWKULAR_SERVER_PORT} \
-Dhawkular.rest.user=${HAWKULAR_AGENT_USER} -Dhawkular.rest.password=${HAWKULAR_AGENT_PASSWORD} \
-Dhawkular.agent.immutable=true -Dhawkular.agent.in-container=true
```

Or would you prefer setting up directly on the YML?
i.e.
```yml
subsystem:
  immutable: "${hawkular.agent.immutable:true}"
  in-container: "${hawkular.agent.in-container:true}"
```

or

```yml
subsystem:
  immutable: true
  in-container: true
```

jmazzitelli:
> Or would you prefer setting up directly on the YML?

I'll let @pilhuhn chime in for his thoughts, but since this is a docker image - I'm not sure we would ever want in-container to be false so setting it the way you have is good. Immutable I'm less sure - but it seems in a docker container we always want immutable to be set to true, too. So, again, the way you have it seems OK. The reason we have those overriding system properties is to override whatever is in the config file - it ensures those values are set to what we want. Because these are docker images, I think we always wants these true which is why I think setting the sysprops directly is OK.

pilhuhn:
Yes, directly put them in the image. We treat them as immutable until further notice.

josejulio:
> Yes, directly put them in the image. We treat them as immutable until further notice.

So the current way is OK?
```Dockerfile
CMD /opt/jboss/wildfly/bin/standalone.sh -b 0.0.0.0 \
-Dhawkular.rest.host=${HAWKULAR_SERVER_PROTOCOL}://${HAWKULAR_SERVER_ADDR}:${HAWKULAR_SERVER_PORT} \
-Dhawkular.rest.user=${HAWKULAR_AGENT_USER} -Dhawkular.rest.password=${HAWKULAR_AGENT_PASSWORD} \
-Dhawkular.agent.immutable=true -Dhawkular.agent.in-container=true
```

jmazzitelli:
> So the current way is OK?

I think he is saying it is OK. When this goes green, we will merge.

docker-dist/Dockerfile-wf-agent
```diff
@@ -0,0 +1,48 @@
...
+  --password '${env.HAWKULAR_AGENT_PASSWORD}' --target-location /opt/jboss/wildfly \
+  --feed-id='${hawkular.agent.machine.id}'
+
+USER root
+RUN chown -R jboss:0 /opt/jboss/wildfly/standalone && \
+    chmod -R ug+rw /opt/jboss/wildfly/standalone && \
+    rm /tmp/installer.jar
+
+USER jboss
+CMD /opt/jboss/wildfly/bin/standalone.sh
```
jmazzitelli, line 47: shouldn't the immutable/in-container flags go here too?

pilhuhn: That should be set in standalone.conf - at least this is how I did it way back

josejulio: ~~Yep, and in the -domain also. Will update the Dockerfiles.~~
Right, they are in the standalone.conf

docker-dist/Dockerfile-wf-agent-domain
```diff
@@ -31,15 +32,15 @@ EXPOSE 8080 9090
...
+  --password '${env.HAWKULAR_AGENT_PASSWORD}' --target-location /opt/jboss/wildfly
 
 USER root
 RUN chown -R jboss:0 /opt/jboss/wildfly/domain && \
     chmod -R ug+rw /opt/jboss/wildfly/domain && \
     rm /tmp/installer.jar
 
-USER JBOSS
+USER jboss
 CMD /opt/jboss/wildfly/bin/domain.sh
```
jmazzitelli, line 46: shouldn't the immutable/in-container flags go here too?

docker-dist/Dockerfile-wf-agent
josejulio: Question is, want me to move them to the Dockerfile or leave them there?

jmazzitelli: oh right! For wildfly agent, its in .conf. Leave them there. I totally forgot. I will merge this is green now.

docker-dist/Dockerfile-wf-agent-domain
jmazzitelli: ignore me - I forgot we have this in the .conf file

jmazzitelli: except for domain mode, it doesn't use standalone.conf. Is this in the proper place for domain mode?

jmazzitelli: We need immutable/in-container flag in a domain.conf unless I am missing something (again) :)

josejulio: See above and refer to this PR https://github.com/hawkular/hawkular-agent/pull/311:)

PR #315, opened 2017-04-07
Hawkular-1208 Notification Mechanism

jshaughn:
- This new notification mechanism provides the agent-side solution to the Jira in question.
- Add new "notification-dmr" meta-data that causes notifications to be sent to the server based on various inventory events.
-- Currently only "resource-added" is supported.
-- Uses services REST API 

Marked as DoNotMerge until we can update to a server src dep.  A slight versioning game is unfortunately going to be needed here due to a circular dep.  Agent depends on server api, server depends on agent extension...  I suppose we could avoid this if we didn't depend on the new api jar and generated explicit JSON in the requests.  Something to consider, I guess.  Although as long as the api doesn't change we'll only have to play this game once.

Also, tests likely need to be added, not sure I can get that in place before PTO.

@jmazzitelli Please review.


jmazzitelli:
to avoid circular dep, we should put API jars in https://github.com/hawkular/hawkular-commons somewhere.

jshaughn:
Updated to use new commons hawkular-client-api jar to remove circular deps. Still marked as DNM due to commons snapshot dep and probabaly need to add an itest.

jmazzitelli:
before we merge, @jshaughn needs to merge my PR https://github.com/jshaughn/hawkular-agent/pull/8 to this PR to fix some things

jshaughn:
I think this is ready to merge.

PR #317, opened 2017-04-12
Adding the automation for pushing the docker images by Travis

jkremser:
It's done only if there is a new release.

~It's basically done, but I still need to test it.~

@jmazzitelli I tried the script on my local account and it successfully pushed the image (https://hub.docker.com/r/jkremser/wildfly-hawkular-agent/tags/)

btw. the `:latest` tag does not mean the last successful build of the master branch (as in your hosa repo), but it's the highest tag. This will cause less surprises.

It can be used also on its own like this:

```bash
OWNER="jkremser" TRAVIS_TAG="0.0.1.Final" DOCKER_USER="jkremser" DOCKER_PASS="***" ./.travis.push.docker.images.sh
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,10 @@ env:
   global:
   - secure: DOP9OifGr7yBDNdkn1Tl58ur4tJzqFaNzVktmX+KgpjDbWtKE3cqyqf1pElyAbSpTm8kVf/UAxUPyCXdRh/W7jj2jB6Tc5kwBCqM8UHjJ6QuZVGnpih4lO6EImV1MDW2+MNWvun7kdUgtRjCWohaYh4pTPAwOk8v7IfeewH1Ud4=
   - secure: IG9j773bZtof+VCK20doXL6HiMnWfQVPD2MhkPc0wc1sXBi3QPatcsWqnOQSsTODyzZQu2yXrRD7cCsXHBFusYxMPIra08mjN4i16b6D4UdtpEhEucotqYCgawGhbv0kW/iRmGKPphAV4pZn+FNiBKJJQzBthBVnzz6WYkYdz7w=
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -47,3 +51,6 @@ after_success:
   then
     ./mvnw -s .travis.maven.settings.xml deploy -DskipTests -Dlicense.skip=true;
   fi
+- if [[ "$TRAVIS_TAG" != ""]] && [[ "${TRAVIS_BRANCH}" = "master" ]] && [[ "${TRAVIS_PULL_REQUEST}" = "false" ]]; then
+    ./.travis.push.docker.images.sh
+  fi
```
jmazzitelli: No good. Looking at travis output I see syntax errors in the docker script at this if stmt. Here's the output:

```
$ if [[ "$TRAVIS_TAG" != ""]] && [[ "${TRAVIS_BRANCH}" = "master" ]] && [[ "${TRAVIS_PULL_REQUEST}" = "false" ]]; then ./.travis.push.docker.images.sh fi
/home/travis/.travis/job_stages: eval: line 53: conditional binary operator expected
/home/travis/.travis/job_stages: eval: line 53: syntax error near `"${TRAVIS_BRANCH}"'
/home/travis/.travis/job_stages: eval: line 53: `if [[ "$TRAVIS_TAG" != ""]] && [[ "${TRAVIS_BRANCH}" = "master" ]] && [[ "${TRAVIS_PULL_REQUEST}" = "false" ]]; then ./.travis.push.docker.images.sh fi '
```

jmazzitelli (changes requested):

.travis.yml
jkremser: it looks better now

jmazzitelli (approved):

PR #318, opened 2017-04-12
HWKAGENT-167 Hawkular should relocate agent dependency classes

cunningt:
https://issues.jboss.org/browse/HWKAGENT-167


jmazzitelli:
@cunningt is this ready to merge or are you going to be doing anything else? What about jboss logging? Can that get relocated or no?

cunningt:
@jmazzitelli I haven't had success yet in relocating jboss logging - still working on that, but I think this pull seems ready to go, and I'll submit another for JBoss Logging if I can fix that with relocation.

Is there anywhere else where the java-agent is pulled in?   I've added a classifier to the artifact that's necessary:

                    <dependency>
                      <groupId>org.hawkular.agent</groupId>
                      <artifactId>hawkular-javaagent</artifactId>
                      <version>${project.version}</version>
                      <classifier>shaded</classifier>
                    </dependency>

and it would be necessary for anything pulling in the artifact.

PR #319, opened 2017-04-12
Use 127.0.0.1 to talk with local wildfly when bind address is 0.0.0.0

josejulio:


josejulio:
@jmazzitelli could you take a look?

jmazzitelli:
LGTM - will merge once travis is green - but I see no reason why this should fail

PR #320, opened 2017-04-18
Merge the new inventory into agent

metlos:
This will not pass until https://github.com/hawkular/hawkular-commons/pull/99 is merged and the code here is updated with the correct srcdeps version of hawkular-commons.

In addition to trivial merge changes, I had to make updates to the new `NotificationDispatcher`. That will need a careful review.


metlos:
@jshaughn could you please review this? In particular the changes to https://github.com/hawkular/hawkular-agent/pull/320/files#diff-8e3b282fb23736ce7992d5b9472435e9

hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/storage/NotificationDispatcher.java
```diff
@@ -63,14 +62,4 @@ public NotificationDispatcher(StorageAdapter storageAdapter, String feedId) {
                     }
                 });
     }
```
jshaughn, line 64: From what I can tell the mechanism will still work.  The difference is only that we'll send a RESOURCE_ADDED notification if the resource is newly discovered OR updated in some way.  Because we already protect against duplicate RESOURCE_ADDED events we should still be OK, despite potentially sending more unnecessary notifications due to updates.  Because we can't distinguish between adds and updates, it precludes our ability to generate explicit update notifications (but note that the previous version of InventoryListener did not provide a resourcedUpdated() method, so we probably haven't lost anything in this regard).

jshaughn (approved):

jmazzitelli:
@metlos can you squash these commits?

jmazzitelli:
PR #325 is the one we are going to use - close this one

PR #321, opened 2017-04-19
Put semicolons here so travis doesn't spit out error

jmazzitelli:
if [ "${cachedBytes}" -gt "${availBytes}" ] ; then echo "Cleaning the cached dirs (${cachedBytes} Bytes) because their size exceeds the free space (${availBytes} Bytes) left on the current partition" rm -Rf "$HOME/.m2" fi
/home/travis/.travis/job_stages: eval: line 54: syntax error: unexpected end of file


evanii:
lgtm, travis.yml + bash = evil

jkremser:
lgtm, yaml + bash = evil

PR #322, opened 2017-04-20
Add missing code to load NotificationType on javaagent

josejulio:
notificationType was always null on javaagent because the method `setName` was never called.
It didn't show any errors because `validate` was never called either.

In this PR I call `validate` method from within `DMRResourceType` and create a new NotificationTypeJsonProperty to handle the serialization and deserialization of `NotificationType`.


josejulio:
@jmazzitelli please review

jmazzitelli:
thanks for finding that bug @josejulio 

I'm going to implement the fix in a different way to avoid having to create that extra json property class. Plus, that abstract stringified class was really to support attributes that can have `${X}` tokens in them which we don't need here. I want to keep the "name" field as a string object here rather than introduce a new object. This notification type may also be extended in the future so I want to keep it simple for now.

Closing this to be replaced with PR #323

josejulio:
Just for completeness:

I was following the pattern from `MetricTypeJsonProperty` and `MeasurementUnitJsonProperty`.
 
`AbstractStringifiedProperty` is not used to support replaces like `${X}` you probably mean `AbstractExpression` (although that one inherits from the first, so maybe that is what you mean).

PR #323, opened 2017-04-20
fix notification pojo

jmazzitelli:
- make sure validate is called on notifications
- add unit test
- clean up the pojo a bit


jmazzitelli:
I hate travis. This all passes on my box. I'm merging.

PR #332, opened 2017-05-01
HWKAGENT-173 remote immutable and in-container attributes

jmazzitelli:


jmazzitelli:
The MiQ stuff will now look at the agent attributes, so we can merge this now @jshaughn. See https://github.com/ManageIQ/manageiq-providers-hawkular/pull/16/files

jshaughn:
+1, the MIQ commit looks good and these removals lgtm.

PR #336, opened 2017-05-04
Increase inventory expiry detection margin

jotak:
... to 1 day


jotak:
@jmazzitelli it looks like I was wrong yesterday, the margin was set to just 5 minutes! So not enough if a single discovery was skipped for whatever reason.

PR #337, opened 2017-05-04
Remove 'Hawkular' from local-dmr resource-type-sets.

josejulio:
This set is to read the hawkular wildfly agent (subsystem).
Since we are using javaagent, its not needed here.


jmazzitelli:
What about the EAP6 feature pack? See https://github.com/hawkular/hawkular-agent/blob/master/eap6-support/hawkular-javaagent-wildfly-feature-pack-eap6/src/main/resources/featurepack/content/standalone/configuration/hawkular-javaagent-config.yaml#L1252

josejulio:
Totally forgot EAP6 feature pack existed :+1:

josejulio:
Updated, thanks

PR #340, opened 2017-05-06
HWKAGENT-176 handle timeout issue when performing javaagent cli jmx Stop

jshaughn:
Add some support for ignoring responses for a command that won't respond.

@jmazzitelli, I played around with a variety of ways to try and handle this issue and this one seems to work while not seeming too terrible to me.  Take a look, see what you think.  The issue happened to me when I wrote an itest for testing the avail backfiller in services.  I used CommandCli to issue a "Stop" JMX operation and it hung my test for 5 minutes before the readTimeout happened.  Using a short timeout is another approach, but it still causes a timeout, which is a little hacky and looks bad in the log.  This approach leaves the log clean, which is a benefit.

### DIFF
--- a/hawkular-agent-command-cli/src/main/java/org/hawkular/agent/commandcli/CommandCli.java
+++ b/hawkular-agent-command-cli/src/main/java/org/hawkular/agent/commandcli/CommandCli.java
@@ -104,6 +104,7 @@ public class CommandCli {
         String jsonRequest;
         File binaryDataFile;
 
+        @Override
         public String toString() {
             StringBuilder str = new StringBuilder("CLI Configuration:\n");
             str.append("Server URL:        ").append(this.serverUrl).append("\n");
@@ -115,6 +116,11 @@ public class CommandCli {
             str.append("Binary Data File:  ").append(this.binaryDataFile).append("\n");
             str.append("JSON Request:      ").append(this.jsonRequest).append("\n");
             return str.toString();
+        }
+
+        public boolean willRespond() {
+            boolean willRespond = !jsonRequest.toLowerCase().contains(":\"stop\"");
+            return willRespond;
         }
     }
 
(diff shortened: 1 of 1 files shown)


hawkular-javaagent/src/main/java/org/hawkular/agent/javaagent/JavaAgentEngine.java
```diff
@@ -235,7 +235,14 @@ public void start() {
...
     public void stop() {
-        stopHawkularAgent();
+        // Ensure the JMX stop command returns successfully by doing this in a thread.
+        Runnable r = new Runnable() {
+            @Override
+            public void run() {
+                stopHawkularAgent();
+            }
+        };
+        new Thread(r).start();
```
jmazzitelli: Give a name to this thread, so if it hangs or blows up, the stack trace will give us a clue who called it. i.e:

```
   new Thread(r, "Hawkular Java Agent Stop Thread").start();
```

I also need to think about this some more before merging since we are now making the stop() method asynchronous. We have to check to see what kinds of thread safety/concurrency issues this might cause (especially in cases where the agent is restarted, for example, and the stop is followed quickly with start. We need to make sure nothing in start is initiated before this stop thread is actually kicked off and finishes). In other words, making stop method asynchronous could have some side affects that are "bad" without some synchronization/locking put in place.

jshaughn: You know, let me try this without spawning the thread, that may be an unnecessary holdover from earlier approaches.  I think the main bit now is just not waiting for a response.  I'll test and let you know...

jshaughn:
OK, good news, the thread is not necessary, so now it's just the bit that ignores the response.  So, only question now is whether you mind that "willRespond()" has hardcoded knowledge of which command won't respond, or whether that should somehow be configurable.

jmazzitelli:
I think this is fine. @jshaughn  This helps workaround the problem with your tests?

jshaughn:
@jmazzitelli Yes, when this is merged I can apply a src-dep to services and then I can generate the PR for the new backfill itest, which is the test that found this issue.

PR #342, opened 2017-05-09
Logs agent version on startup

josejulio:
On a Hawkular javaagent it looks like this:
```
14:18:22,592 INFO  ...: Hawkular Agent: Javaagent version 1.0.0.Final-SNAPSHOT
```

On a Hawkular wildfly agent it looks like this:
```
14:11:21,450 INFO  ...: Hawkular Agent: WildFly Agent Subsystem version 1.0.0.Final-SNAPSHOT
```

### DIFF
--- a/hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/service/AgentCoreEngine.java
+++ b/hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/service/AgentCoreEngine.java
@@ -225,6 +225,11 @@ public abstract class AgentCoreEngine {
         }
 
         try {
+            Package agentPackage = this.getClass().getPackage();
+            if (agentPackage != null) {
+                log.infoTypeAndVersion(agentPackage.getImplementationTitle(), agentPackage.getImplementationVersion());
+            }
+
             log.infoStarting();
 
             // determine the configuration to use
--- a/hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/log/MsgLogger.java
+++ b/hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/log/MsgLogger.java
@@ -253,4 +253,8 @@ public interface MsgLogger extends BasicLogger {
     @LogMessage(level = Level.ERROR)
     @Message(id = 10083, value = "Failed to store notification: %s")
     void errorFailedToStoreNotification(@Cause Throwable t, String data);
+
+    @LogMessage(level = Level.INFO)
+    @Message(id = 10084, value = "%s version %s")
+    void infoTypeAndVersion(String type, String version);
 }


josejulio:
@jmazzitelli is this what you wanted?

hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/util/Util.java
```diff
@@ -389,4 +392,17 @@ public static String extractDockerContainerIdFromString(String s) {
...
+        String filename = clazz.getName().replace('.', '/') + ".class";
+        String classPath = clazz.getClassLoader().getResource(filename).toString();
+        if (!classPath.startsWith("jar")) {
+            return null;
+        }
+        String manifestPath = classPath.substring(0, classPath.lastIndexOf('!') + 1) + File.separator + JarFile.MANIFEST_NAME;
+        InputStream stream = new URL(manifestPath).openStream();
+        Manifest manifest = new Manifest(stream);
+        stream.close();
+        return manifest;
```
jmazzitelli: This seems overkill. Can we just remove this Util method and in AgentCoreEngine simply do:

```
log.infoTypeAndVersion(this.getClass().getPackage().getImplementationTitle(),
    this.getClass().getPackage().getImplementationVersion());
```

hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/service/AgentCoreEngine.java
```diff
@@ -225,6 +227,14 @@ public void startHawkularAgent(AgentCoreEngineConfiguration newConfiguration) {
...
 
         try {
+            Manifest agentManifest = Util.manifestFromClass(this.getClass());
+            if (agentManifest != null) {
+                Attributes manifestAttributes = agentManifest.getMainAttributes();
+                String agentType = manifestAttributes.getValue("Implementation-Title");
+                String agentVersion = manifestAttributes.getValue("Implementation-Version");
+                log.infoTypeAndVersion(agentType, agentVersion);
+            }
+
```
jmazzitelli: Can we just remove all of this and replace it with one line like this:

```
log.infoTypeAndVersion(this.getClass().getPackage().getImplementationTitle(),
    this.getClass().getPackage().getImplementationVersion());
```

jmazzitelli:
@josejulio I'm not sure if getPackage() returns the manifest information - but I think it does. I haven't tested it, but I'm pretty sure I have done something like this in the past.

hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/util/Util.java
josejulio: Sure. I didn't know about those.

josejulio:
I'll test and update this.

josejulio:
Just updated it.
Tested on hawkular-javaagent and hawkular-wildfly-agent.

PR #343, opened 2017-05-10
HWKAGENT-179 - do not submit notifications when not in full hawkular mode

jmazzitelli:


jshaughn:
Right, the notification stuff uses the hservices rest api, so it should be ignored in metrics mode. LGTM.

PR #344, opened 2017-05-23
HWKAGENT-180 be able to configure the local-dmr protocol.

jmazzitelli:
When attaching to EAP6 this has to be set to "remote"


jshaughn:
LGTM

PR #345, opened 2017-06-08
HWKAGENT-181 change the avail regex

jmazzitelli:
UP avail includes servers in reload-required and restart-required states


josejulio:
Could you also update this one?

https://github.com/jmazzitelli/hawkular-agent/blob/4fafdbb132386dbb79ae4381fbc32ede9be24238/eap6-support/hawkular-javaagent-wildfly-feature-pack-eap6/src/main/resources/featurepack/content/standalone/configuration/hawkular-javaagent-config.yaml#L517

PR #347, opened 2017-07-04
HWKAGENT-182 Adds an AvailChange notification when a watched resource…

josejulio:
… change any of its availabilites

This PR adds an `AvailManager` to keep track of last resources availability value.
An `AvailEvent` is sent by `EndpointService` whenever a change on availability is detected. 
`NotificationDispatcher` now also listens to these events. A notification is fired when the resource has `AVAIL_CHANGE` on its notifications list with the `resourceType`, `resourcePath`, availType` and `newAvail`.

Depends on https://github.com/hawkular/hawkular-commons/pull/108

### DIFF
--- a/hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/inventory/AvailManager.java
+++ b/hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/inventory/AvailManager.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2015-2017 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.hawkular.agent.monitor.inventory;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.hawkular.agent.monitor.api.Avail;
+
+/**
+ * Holds a map between resource ids and avails. Provides methods for retrieve data and update the data in the map.
+ *
+ * @author <a href="https://github.com/josejulio">Josejulio Martínez</a>
+ */
+public final class AvailManager<L> {
+
+    public static class AddResult<L> {
+        public enum Effect {
+            MODIFIED, UNCHANGED
+        }
+        private final MeasurementInstance<L, AvailType<L>> measurementInstance;
+        private final Avail avail;
+        private final Effect effect;
+
+        public AddResult(MeasurementInstance<L, AvailType<L>> measurementInstance, Avail avail, Effect effect) {
+            this.measurementInstance = measurementInstance;
+            this.avail = avail;
+            this.effect = effect;
+        }
+
+        public MeasurementInstance<L, AvailType<L>> getMeasurementInstance() {
+            return measurementInstance;
+        }
+
+        public Avail getAvail() {
+            return avail;
+        }
+
+        public Effect getEffect() {
+            return effect;
+        }
+    }
+
+    private Map<MeasurementInstance<L, AvailType<L>>, Avail> storage = new HashMap<>();
+
+    public AddResult<L> addAvail(MeasurementInstance<L, AvailType<L>> measurementInstance, Avail avail) {
+        AddResult<L> result;
+        Avail previousAvail = storage.get(measurementInstance);
+        if (previousAvail == null || avail == previousAvail) {
+            result = new AddResult<>(measurementInstance, avail, AddResult.Effect.UNCHANGED);
+        } else {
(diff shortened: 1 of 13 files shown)


josejulio:

hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/storage/NotificationDispatcher.java
```diff
@@ -62,4 +64,33 @@ public NotificationDispatcher(StorageAdapter storageAdapter, String feedId) {
...
+                    try {
+                        NotificationPayloadBuilder b = storageAdapter.createNotificationPayloadBuilder();
+                        b.addNotificationType(NotificationType.AVAIL_CHANGE);
+                        b.addProperty("resourceType", mi.getResource().getResourceType().getName().getNameString());
+                        b.addProperty("resourcePath", cp.toString());
+                        b.addProperty("availType", mi.getType().getName().getNameString());
+                        b.addProperty("newAvail", event.getChanged().get(mi).name());
+                        storageAdapter.store(b, 0);
+                    } catch (Exception e) {
+
```
jmazzitelli: should we log an error message here rather than silently ignore it?

jmazzitelli:
as I understand this PR, what is happening is the agent is now storing all avail metrics that it collects. When a new avail is detected (i.e. it wasn't collected before during the agent's lifetime) or when the agent detects a new avail value changes a previously collected avail value AND that resource has "avail-changed" notification metadata defined on it, a notification message is sent up to the hawkular server with information about the resource that was changed and what the new avail is.

If this understanding is correct, this looks OK to me.

Travis is red, just need to make sure this is really passing before we consider merging.

hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/storage/NotificationDispatcher.java
josejulio: Sure, not sure why it wasn't before.

jshaughn:
>  When a new avail is detected (i.e. it wasn't collected before during the agent's lifetime) ...

I don't think we want to send a notification for the first avail, then we'd get events for every first avail collected after startup (for relevant resources).

josejulio:
> When a new avail is detected (i.e. it wasn't collected before during the agent's lifetime)

It will mark them as `UNCHANGED` if there is no previousAvail (is the first) or the avail is the same as the previous.
```java 
previousAvail == null || avail == previousAvail
 ```

File: hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/inventory/AvailManager.java [hunk 2]
```diff
@@ -0,0 +1,74 @@
...
+        }
+    }
+
+    private Map<MeasurementInstance<L, AvailType<L>>, Avail> storage = new HashMap<>();
+
+    public AddResult<L> addAvail(MeasurementInstance<L, AvailType<L>> measurementInstance, Avail avail) {
+        Avail previousAvail = storage.get(measurementInstance);
+        AddResult.Effect effect;
+        if (previousAvail == null) {
+            effect = AddResult.Effect.STARTING;
```
josejulio, line 64: I haven't (yet) tested it, but what do you think of this idea @jshaughn ?

It will inform on its first avail, and services will handle it as a normal avail-change.

jshaughn: Maybe I'm forgetting something but why would we need to track and send STARTING events for all resources?  The only thing we care about is when a server starts because it is implicitly an avail change (because we assume it goes up and down with the agent).   It doesn't look to me like we need this extra tracking, extra messaging, and extra server-side handling.  Let me know if I am not understanding this correctly.

josejulio: Just trying to be generalize and avoid being too particular on agent side (as most of the work is general and driven by configuration).
This would indeed make the agent to send once (on first avail of a resource) the avail_starting for all resources to `NotificationDispatcher`.
However it will only send notifications for resources that have the "avail-starting" notification on their notification-dmr list (currently `WildFly Server`, `Host Controller`, `Domain Host` and `Domain WildFly Server Controller`) see https://github.com/hawkular/hawkular-agent/pull/347/files#diff-8e3b282fb23736ce7992d5b9472435e9R87

That said, I can still put code on the AvailManager to only do that for certain resource types, as (conveniently) the metrics know which resource they are tracking.

jshaughn: Ah, I see.  OK, this is fine then, as long as we are only sending STARTING messages to the server for the configured resource types.  Then, on the server we can map that to an avail change event or, I suppose just map it directly to a Starting event.

josejulio: I'm currently processing it as a AVAIL_CHANGE on services https://github.com/hawkular/hawkular-services/pull/169/files#diff-1df0cac565a159c01b113e2ffb7dcaa1R95

jshaughn: That's fine by me.  If at some point a PM wants to see STARTED then we already have it in place.

jshaughn:
I merged the commons PR, do you want to do a release of commons or use a src-dep in the agent pom?

josejulio:
I suppose is OK to do use src-dep and release later.

josejulio:
Updated commons with src-dep and tests are passing now.

jshaughn:
LGTM

PR #348, opened 2017-07-17
Sets correct relative-to path depending if using standalone/domain

josejulio:
When installing on domain mode, sets the correct `relative-to` path.

Host doesn't know anything about `jboss.server.config.dir`
```
[domain@localhost:9990 /] /host=master/:resolve-expression(expression=${jboss.server.config.dir})
{
    "outcome" => "failed",
    "failure-description" => "WFLYCTL0211: Cannot resolve expression '${jboss.server.config.dir}'",
    "rolled-back" => true
}
```

But is well aware of `jboss.domain.config.dir`
```
[domain@localhost:9990 /] /host=master/:resolve-expression(expression=${jboss.domain.config.dir})
{
    "outcome" => "success",
    "result" => "/home/josejulio/Aplicaciones/redhat/jboss-eap-7.0/domain/configuration"
}
```

### DIFF
--- a/hawkular-wildfly-agent-installer/src/main/java/org/hawkular/wildfly/agent/installer/AgentInstaller.java
+++ b/hawkular-wildfly-agent-installer/src/main/java/org/hawkular/wildfly/agent/installer/AgentInstaller.java
@@ -304,7 +304,7 @@ public class AgentInstaller {
                     Files.copy(Paths.get(keystoreSrcFile.getAbsolutePath()), keystoreDst);
                 }
 
-                String securityRealm = createSecurityRealm(keystoreSrcFile.getName(), keystorePass);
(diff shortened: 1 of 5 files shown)


josejulio:
@jmazzitelli could you take a look?

jmazzitelli:
LGTM, but I wonder how this ever worked before? 

@josejulio Before I merge, do you know why we need to do this now? How did this work before? At least, I'm pretty sure this did work before.

josejulio:
I'm not sure that installing (using the installer) into host.xml and providing a keystore worked before. 
I will check if we did that before.

josejulio:
This [1] seems to be the only recent modification of that code. 
I don't think it broke it though.

[1] https://github.com/hawkular/hawkular-agent/commit/1b3f887945475f05c9b9bc4c079ac1b810c3c045#diff-61ed95d87c2b546931ac6edb549e895aL393

jmazzitelli:
LGTM

PR #349, opened 2017-07-19
Adding JDR operation for domain servers and support to delete JDR files

israel-hdez:
* Changing configuration files to be able to generate JDR reports
  servers running in domain mode.
* Changes to delete JDR report files after [Export JDR] operation is
  complete.

### DIFF
--- a/eap6-support/hawkular-javaagent-wildfly-feature-pack-eap6/src/main/resources/featurepack/content/standalone/configuration/hawkular-javaagent-config.yaml
+++ b/eap6-support/hawkular-javaagent-wildfly-feature-pack-eap6/src/main/resources/featurepack/content/standalone/configuration/hawkular-javaagent-config.yaml
@@ -562,7 +562,6 @@ resource-type-set-dmr:
     operation-dmr:
     - name: JDR
       internal-name: generate-jdr-report
-      modifies:      true
       path:          /subsystem=jdr
     - name: Reload
       internal-name: reload
@@ -750,6 +749,10 @@ resource-type-set-dmr:
     - name: Base Directory
       attribute: base-dir
       path:      /core-service=server-environment
+    operation-dmr:
+    - name: JDR
+      internal-name: generate-jdr-report
+      path:          /subsystem=jdr
 
   - name: Domain WildFly Server Controller
     resource-name-template: "Domain WildFly Server Controller [%-]"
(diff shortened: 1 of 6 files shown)


eap6-support/hawkular-javaagent-wildfly-feature-pack-eap6/src/main/resources/featurepack/content/standalone/configuration/hawkular-javaagent-config.yaml
```diff
@@ -750,6 +750,11 @@ resource-type-set-dmr:
     - name: Base Directory
       attribute: base-dir
       path:      /core-service=server-environment
+    operation-dmr:
+    - name: JDR
+      internal-name: generate-jdr-report
+      modifies:      true
```
jmazzitelli: modifies=true should probably be removed here as well as in the other place in hawkular-wildfly-agent-feature-pack/src/main/resources/subsystem-templates/hawkular-wildfly-agent.xml.

Note that you do not have modifies in the subsystem XML config (so the two configs - for javaagent and subsystem agent - are different now).

modifies=true means the operation will change something in the managed resource (perhaps a configuration setting or some runtime behavior). When this is the case, the agent will disallow this operation from running when run in a container or when it is "immutable".

Asking for a JDR report (AFAIU) is simply a read-only function and returns a debug report. If that is true, "modifies" should probably be false to allow someone to ask for a JDR report even if the agent is running in immutable or in-container mode. Otherwise, this modifies=true will not allow anyone to get debug information via JDR when running  while immutable/in-container.

israel-hdez: I think you are right, it should be a read-only function.
I removed that property from Domain Wildflies and also for standalone ones.

jmazzitelli:
```
[ERROR] /home/travis/build/hawkular/hawkular-agent/hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/cmd/ExportJdrCommand.java:[70,44] cannot find symbol
  symbol:   method isDeleteImmediately()
  location: variable request of type org.hawkular.cmdgw.api.ExportJdrRequest
```

Need to pull in the new cmdgw commons library.

israel-hdez:
Now it compiles, but I don't know why an integration test didn't pass.

israel-hdez:
It's passing now...

jmazzitelli:
LGTM

PR #354, opened 2017-09-18
When using javaagent, add system properties to JAVA_OPTS.

josejulio:
Every param specified after standalone.sh is sent to main method args
e.g. main(String ... args)

The reason it works sometimes, is because wildfly takes the -D params
and [sets](https://github.com/wildfly/wildfly/blob/master/appclient/src/main/java/org/jboss/as/appclient/subsystem/Main.java#L245-L256) them. When using javaagent, we run the startup code in a
thread. So depending on your luck (most of the times bad luck fo me)
your agent wouldn't *catch* the those params and you would end up
wondering why.

On an agent as subsystem it worked well because wildfly starts
the subsystem after it set everything up.


josejulio:

PR #355, opened 2017-10-02
Inventory: build modified resources for import

jotak:
Here all resources under a modified root are imported. Future improvement could be to import only created or modified children.

### DIFF
--- a/hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/service/AgentCoreEngine.java
+++ b/hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/service/AgentCoreEngine.java
@@ -573,7 +573,6 @@ public abstract class AgentCoreEngine {
         this.storageAdapter.initialize(
                 feedId,
                 configuration.getStorageAdapter(),
-                configuration.getGlobalConfiguration().getAutoDiscoveryScanPeriodSeconds(),
                 diagnostics,
                 httpClientBuilder);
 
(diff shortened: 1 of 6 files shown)


hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/service/AgentCoreEngine.java
```diff
@@ -573,7 +573,6 @@ private void startStorageAdapter() throws Exception {
         this.storageAdapter.initialize(
                 feedId,
                 configuration.getStorageAdapter(),
-                configuration.getGlobalConfiguration().getAutoDiscoveryScanPeriodSeconds(),
```
jmazzitelli, line 576: why are you getting rid of this? The agent needs to scan for new discovery periodically throughout its lifetime. We can't have it just do discovery at startup and never do it again. It has to be done every X seconds so we can detect resources coming up and going away (and tell inventory when that happens).

jotak: Iirc I added this to anticipate the metrics TTL, but there's no reason to keep it passed to AsyncInventoryStorage constructor. So I just removed it from there and mechanically removed in call hierarchy, but it's still there for the discovery job.

jmazzitelli: oh, i see. It isn't removed, just not passed to this object. I get it now.

jotak:
@jmazzitelli i've simplified by pushing directly the impacted resources, no more whole root tree sync :)
That is made possible with inventory using parentId now, same as the agent does.

PR #361, opened 2017-10-04
put feed id in resource ids, but not resource type ids

jmazzitelli:
resource type ids are back to the way they were, which is just the type name


jotak:
LGTM

hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/storage/AsyncInventoryStorage.java
```diff
@@ -166,7 +166,7 @@ private String getInventoryId(NamedObject no) {
         } else {
             id = no.getID().getIDString();
         }
-        return feedId + "~" + id;
+        return id;
```
jotak, line 169: So resources always have no.getID() != null, and resource type always have it == null?

jmazzitelli: yes, that is the way things work currently.

PR #366, opened 2017-10-07
support the case when no resource type set names are specified in config

jmazzitelli:


jmazzitelli:
We need this when we start supporting inventory metadata config files served by the server.

jshaughn:
LGTM assuming we still need all of the remote mgmt support.

PR #370, opened 2017-10-12
Delete several resources at once

jotak:


hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/storage/AsyncInventoryStorage.java
```diff
@@ -254,9 +258,12 @@ private void sendImportRestRequest(Inventory importData) throws Exception {
...
     }
 
-    private void sendDeleteResourceRestRequest(ID resourceId) throws Exception {
+    private void sendDeleteResourcesRestRequest(List<String> resourceId) throws Exception {
+        String ids = resourceId.stream()
+                .map(Util::urlEncode)
+                .collect(Collectors.joining(","));
         StringBuilder url = Util.getContextUrlString(config.getUrl(), config.getInventoryContext())
-                .append("resources/").append(Util.urlEncode(resourceId.getIDString()));
+                .append("resources?ids=").append(ids);
```
jotak: I'm suddenly having a doubt, if we must write `?ids=A,B,C` or `?ids=A&ids=B&ids=C` ... checking if it's covered by test

jotak:
I triggered a new build, but the test `testRemoveJdbcDriver` failed whereas it passes on my machine. I'm trying to understand what could be some non-deterministic cause here...

jotak:
Ok... it passes, but it's a bit concerning

PR #373, opened 2017-10-16
temp fix to support h-metrics

jmazzitelli:


jmazzitelli:
asking either @jotak or @jshaughn  to review

hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/storage/AsyncInventoryStorage.java
```diff
@@ -201,6 +201,10 @@ private String getInventoryId(NamedObject no) {
             mb.unit(MetricUnit.valueOf(metricUnits.name()));
         }
         m.getProperties().forEach((k, v) -> mb.property(k, v.toString()));
+
+        // TODO remove these once h-metrics integration is removed
+        mb.property("hawkular.metric.type", m.getType().getMetricType().name());
+        mb.property("hawkular.metric.typeId", m.getType().getID().getIDString());
```
jotak, line 207: That's what I wanted to check with you on irc: we don't want the typeId here, but the metric Id (the string like `MT~feed~resource~etc.` I don't remember exactly the pattern)

jotak: And so, the key would be "hawkular.metric.id"

hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/storage/AsyncInventoryStorage.java
```diff
@@ -201,6 +201,11 @@ private String getInventoryId(NamedObject no) {
...
+        mb.property("hawkular.metric.id", m.getID().getIDString());
```
jmazzitelli, line 208: added "hawkular.metric.id" - good @jotak ?

jshaughn: This is good, think, but I also think @jotak is saying that "hawkular.metric.typeId" can go away.  We can wait for his approval or comment before merging.

jmazzitelli: I just left type ID in there just in case it could be useful. We are removing all of these soon anyway, so I figured leave it in case type ID is useful and it'll go away eventually.

jotak: LGTM

PR #375, opened 2017-10-19
HAWKULAR-1280: update the agent's docker/openshift configuration

mwringe:


jmazzitelli:
I can't figure out why that remove JDBC driver fails sometimes. It works on my box. another travis oddity. I would say ignore that for now. that failure has nothing to do with this PR anyway.

jotak:
@jmazzitelli isn't it a matter of immutable container? I think on the ruby gem, the tests of adding/removing jdbc drivers are skipped when immutable is true.

PR #380, opened 2017-10-26
Hawkular 1259

jmazzitelli:


jshaughn:
Adding DNM until we have a green light on the merge/branching plan. Also, as in services, we may want to make a release of commons and then update this to use the release dep, not a src-dep.

jmazzitelli:
I still don't understand that failure in travis. However, ignore it. Because the next set of changes we introduce (removing h-metrics) will mean this test should again pass.

PR #381, opened 2017-10-26
HAWKULAR-1308 - Loading sensitive information as envs instead of system properties

josejulio:

PR #382, opened 2017-10-27
HAWKULAR-1289 - Build a single docker image that supports both standalone and domain

josejulio:
By setting env HAWKULAR_MODE one can run it in standalone or domain.
 If HAWKULAR_MODE is not set, it defaults to standalone.
```bash
   docker run -e HAWKULAR_MODE=standalone hawkular/wildfly-hawkular-javaagent
   docker run -e HAWKULAR_MODE=domain hawkular/wildfly-hawkular-javaagent
```

Depends on #381

### DIFF
--- a/docker-dist/Dockerfile
+++ b/docker-dist/Dockerfile
@@ -15,28 +15,32 @@
 # limitations under the License.
 #
 
-FROM jboss/wildfly:10.1.0.Final
+FROM jboss/wildfly:11.0.0.Final
 
 MAINTAINER Hawkular project <hawkular-dev@lists.jboss.org>
 
 # ADD test-simple.war /opt/jboss/wildfly/standalone/deployments/
 COPY target/hawkular-javaagent.jar $JBOSS_HOME/bin/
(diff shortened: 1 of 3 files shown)


hawkular-javaagent-wildfly-feature-pack/src/main/resources/featurepack/content/standalone/configuration/hawkular-javaagent-config.yaml
```diff
@@ -33,9 +33,9 @@ storage-adapter:
   type:      HAWKULAR
   tenant-id: "${hawkular.rest.tenantId:hawkular}"
   feed-id:   "${hawkular.rest.feedId:autogenerate}"
-  url:       "${hawkular.rest.host:http://hawkular-server:8080}"
-  username:  "${hawkular.rest.user}"
-  password:  "${hawkular.rest.password}"
+  url:       "${hawkular.rest.url:http://hawkular-server:8080}"
```
jmazzitelli: there are other places where this hawkualr.rest.host is defined that also needs to change. Since this is independent of what this PR is really for, just revert this line back to using hawkualr.rest.host and we'll fix that later.

PR #386, opened 2017-10-28
Domain config fix

jshaughn:


jmazzitelli:
we'll merge this if we have to. for now, don't do it yet. these changes will be moot once hawkular-1275 makes its way here (which is going to be soon)

jmazzitelli:
blah - we need this so services itests can pass. merging now...

PR #387, opened 2017-10-28
Hawkular 1275

jmazzitelli:


jmazzitelli:
this cleans up a bunch of stuff in itests and moves the p8s integration forward.

PR #397, opened 2017-10-31
register metrics endpoint to server

jmazzitelli:


jmazzitelli:
we won't have the agent do this explicitly - the inventory will do it for the agent

PR #399, opened 2017-10-31
add metrics endpoint to agent config

jmazzitelli:
The resource config properties on agent resource now has a new "Metrics Endpoint" value so you can figure out where to scrape the metrics.

```
{
  "id": "mazzlap~Local JMX~org.hawkular:type=hawkular-javaagent",
  "name": "Hawkular Java Agent",
  "feedId": "mazzlap",
  "type": {
    "id": "Hawkular Java Agent",
    ...
    "properties": {}
  },
...
  "config": {
    "Immutable": "false",
    "Metrics Endpoint": "127.0.0.1:9779",   <-- THE NEW CONFIG
    "In Container": "false"
  }
}
```

### DIFF
--- a/eap6-support/hawkular-javaagent-wildfly-feature-pack-eap6/src/main/resources/featurepack/content/standalone/configuration/hawkular-javaagent-config.yaml
+++ b/eap6-support/hawkular-javaagent-wildfly-feature-pack-eap6/src/main/resources/featurepack/content/standalone/configuration/hawkular-javaagent-config.yaml
@@ -1175,6 +1175,8 @@ resource-type-set-jmx:
       attribute: Immutable
     - name: In Container
       attribute: InContainer
+    - name: Metrics Endpoint
+      attribute: MetricsEndpoint
     operation-jmx:
     - name: Start
       internal-name: start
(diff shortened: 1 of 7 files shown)


jshaughn:
Just a side comment, why use keys with capital letters and spaces?  I guess it doesn't matter, but it seems weird.

jmazzitelli:
> Just a side comment, why use keys with capital letters and spaces? I guess it doesn't matter, but it seems weird.

These are actually the "names" ( used as display names in MiQ) of the configuration properties - see here for example of the analogous WildFly Server config props:

https://github.com/hawkular/hawkular-agent/blob/hawkular-1275/hawkular-javaagent-wildfly-feature-pack/src/main/resources/featurepack/content/standalone/configuration/hawkular-javaagent-config.yaml#L531-L554

Its the same for the agent as you see here (I just added this new one to the list):

https://github.com/hawkular/hawkular-agent/blob/hawkular-1275/hawkular-javaagent-wildfly-feature-pack/src/main/resources/featurepack/content/standalone/configuration/hawkular-javaagent-config.yaml#L1302-L1305

PR #400, opened 2017-11-01
HAWKULAR-1313 - fix the race condition when removing resources via cmdgw command

jmazzitelli:


jshaughn:
This looks OK to me.  I guess now we write lock while the discovery takes place, in a thread.  This prevents concurrent discovery and makes discovery and removals mutually exclusive.  Would it make any sense to build a new inventory during discovery, instead of updating the existing one, to reduce potential blocking?  Probably not, and it would increase memory consumption.

jmazzitelli:
> Would it make any sense to build a new inventory during discovery, instead
> of updating the existing one, to reduce potential blocking? Probably not,
> and it would increase memory consumption.

No, I don't think we need it. Right, memory would increase if we did that, and updating existing allows us to emit events when things change in the existing inventory. The only issue we were having is that discovery method and the remote method were concurrent and we needed them not to be - that's what this PR fixes. Copying into a new inventory wouldn't have helped here anyway - it wasn't that we were overwriting something in inventory concurrently, it was we were removing things while a scan was in progress and that scan happened to have already discovered the the removed resource earlier. It wasn't even concurrent access that was the problem - it was an ordering issue that this lock solution now fixes.

jshaughn:
OK, this looks good to me, I'll merge.   @lucasponce you may still want to review since you did the investigation into the issue.

lucasponce:
Yes, I ran some tests and in 30 batch executions issue is not visible, so fix works.

jotak:
I also ran the tests a couple of times (not as many as Lucas) and works for me, too

PR #401, opened 2017-11-01
[HAWKULAR-1316] Added info on docs about oc 3.7 command

gbaufake:
I could not use the the command to generate secrets (` oc create secret generic hawkular-javaagent-example-ca --from-file=hawkular-services-ca.crt=/opt/hawkular/hawkular-services-public.pem`) because it is specific from oc 3.7+. 

I think it should be good to advise the users to have that version to use the command


gbaufake:
@mwringe Can you review it?

mwringe (approved):

PR #406, opened 2017-11-06
support our platform metrics again - we now wrap them with MBeans

jmazzitelli:
this is for https://issues.jboss.org/browse/HAWKULAR-1321


jmazzitelli:
this requires https://github.com/hawkular/hawkular-commons/pull/186

jshaughn:
LGTM.  I don't think this needs a src-dep bump for commons but I'll wait on the merge just to have @jmazzitelli confirm.

jmazzitelli:
doesn't need srcdep - the only change in commons was to a file included in the inventory war - doesn't affect the agent (but h-services does need the new commons)

PR #411, opened 2017-11-09
add type version API to the core config.

jmazzitelli:


jmazzitelli:
this is just an addition that goes with the changes lucas added. There is a type-version attribute in the yaml config, we just need to pass that to the core agent config object so the agent can access the type version in other places where the actual yaml object is not available.

PR #412, opened 2017-11-11
support server-side agent config files

jmazzitelli:
This is a way we can support server-side config files for agents to download. There is an associated PR for commons - https://github.com/hawkular/hawkular-commons/pull/191

Do not merge yet. This work is not slated for hawkular-1275.


jmazzitelli:
as per off-line discussion, we are free to merge this into hawkular-1275. removing DNM label.

but do not merge until green. I need to update the pom to remove the snapshot dep on commons and put the correct dep in here. then it should go green.

jshaughn:
LGTM

PR #413, opened 2017-11-13
adds a "version" JMX attribute to the agent mbean

jmazzitelli:
We will need this in the future so we can collect it as a resource property later.


jshaughn:
Seems fine. Waiting to go green...

PR #415, opened 2017-11-16
add metric expression

jmazzitelli:


hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/inventory/MeasurementInstance.java
```diff
@@ -75,4 +73,45 @@ public void setMetricLabels(Map<String, String> metricLabels) {
...
+                    metricString.append(comma)
+                            .append(label.getKey())
+                            .append("=\"")
+                            .append(label.getValue())
+                            .append("\"");
+                    comma = ",";
+                }
+                metricString.append("}");
+            }
+            expr = expr.replaceAll("\\$metric", metricString.toString());
```
jshaughn: I think you should use expr.replace() as opposed to replaceAll().  No need to deal with regex here.

jmazzitelli: I think thinking expressions can list it twice. Is there no way a prom query can specify the metric more than once in an expression?

jmazzitelli: fixed.. didn't realize replace does them all

jshaughn:
closing in favor of https://github.com/hawkular/hawkular-agent/pull/416

PR #420, opened 2017-11-21
clean up yaml

jmazzitelli:
This is a PR that doesn't change any functionality - but it does clean up the yaml config files. There are lots of trivial changes here (we had some weird spacing/indentation in many values and quoted things we didn't need to) - just doing this all at once to make the yaml files more consistent and will avoid irrelevant diffs later if this gets fixed little by little. Just do it all now in this PR. Similar PR was done to the server-side configs in commons.


jmazzitelli:
see also https://github.com/hawkular/hawkular-commons/pull/196

PR #422, opened 2017-11-28
generate feed ID using some wildfly settings

jmazzitelli:
When running in wildfly (domain or standalone mode), we'll  want to use some wildfly things for our feed ID by default.


jshaughn:
Wunderbar.

PR #423, opened 2017-12-01
more work to get domain mode to work better

jmazzitelli:


jmazzitelli:
we first must merge https://github.com/hawkular/hawkular-commons/pull/202 and then bump this PR's srcdep on commons before merging

jshaughn:
There is a lot of mojo magic here with parsing trickery and such, but it still seems like a good solution to the problem, and not really overly complicated.  The inline documentation is good but [outside of this PR] we could use some high level documentation about how the agent is operating in the domain scenario.  LGTM.  Will merge if it goes green with the new commons srcdep...

PR #424, opened 2017-12-01
do not abort agent startup if the metrics endpoint fails to start

jmazzitelli:
I think we want the agent to keep going. But we'll log an ugly error to inform the user that the agent can't serve metrics.


jshaughn:
Perhaps at some point we should add an intermittent retry.  But this is good for now.

PR #425, opened 2017-12-12
try to turn on all WildFly statistics if monitoring a WildFly server

jmazzitelli:
this should be ready

### DIFF
--- a/hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/protocol/dmr/StatisticsControl.java
+++ b/hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/protocol/dmr/StatisticsControl.java
@@ -1,0 +1,197 @@
+/*
+ * Copyright 2015-2017 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.hawkular.agent.monitor.protocol.dmr;
+
+import java.util.Collections;
+import java.util.List;
+
+import org.hawkular.agent.monitor.log.AgentLoggers;
+import org.hawkular.agent.monitor.log.MsgLogger;
+import org.hawkular.agent.monitor.util.WildflyCompatibilityUtils;
+import org.hawkular.dmr.api.OperationBuilder;
+import org.hawkular.dmr.api.OperationBuilder.CompositeOperationBuilder;
+import org.hawkular.dmr.api.OperationBuilder.OperationResult;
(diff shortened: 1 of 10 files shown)


jmazzitelli:
This will only work in standalone mode. You cannot change config for slave servers. See Brian's comment here:

https://issues.jboss.org/browse/WFCORE-771?focusedCommentId=13199892#comment-13199892

And I see that error message when my code tries to change the statistics enable flags on slave servers. So that matches what Brian says should happen.

jmazzitelli:
I believe this is ready.

In short, the agent upon start up will now attempt to turn on the different subsystems' statistics enablement flags. This is the default behavior, but you can turn it off by setting "enable-statistics: false" in the local-dmr or remote-dmr section under the managed-server section.

hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/protocol/dmr/StatisticsControl.java
```diff
@@ -0,0 +1,197 @@
...
+            } else {
+                for (String host : hosts) {
+                    List<String> servers = getDomainHostServers(mcc, host);
+                    for (String server : servers) {
+                        // I was trying to be too clever. This would work awesomely if WildFly would just allow it.
+                        // However, this error occurs when you try to set attributes on slave servers:
+                        // "User operations are not permitted to directly update the persistent configuration
+                        // of a server in a managed domain."
+                        // So, for now, comment this out until perhaps in the future WildFly/EAP will allow this.
+                        // This means when in domain mode, the user must manually turn on statistics themselves.
```
jshaughn, line 63: Wouldn't the slaves be enabled if you enable the controller?

jshaughn (approved):
This seems to be in good shape.  I guess my question is, "Why we are doing this?".   And if we enable stats by default does this have any perf effect?

hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/protocol/dmr/StatisticsControl.java
jmazzitelli: no, the slaves configuration here are not affected by changes in the host controller (in fact, for example, there is no /subsystem=ejb3 in the host controller anyway.

jmazzitelli:
> I guess my question is, "Why we are doing this?"

We are doing this because:

1. By default WildFly does not enable statistics and in this state every metric that Prometheus exporter will collect and scrape for these subsystems will always be 0

2. One of our main contributions to middleware management is to provide monitoring (metrics and graphing) - we are rendered mostly useless if a lot of our subsystem metrics are always 0.

3. Because of this, today we tell users they have to enable these stat flags manually. We do provide a cmdgw command for them to execute (StatisticsCommand), but it requires MiQ to send that message and for the user to tell MiQ to send that message. Someone may start Hawkular and not see many metrics for these subsystems and will not know why. This PR makes things more automatic.

> And if we enable stats by default does this have any perf effect?

I will assume it will affect performance but it is unknown if it is a lot or  a little. I suspect it isn't much. But in any case, this is why I provide "enable-statistics" flag in the agent configuration - you can turn this off if you find the performance hit is not worth it. In fact, by default I set this to true, we could set the default to "false" and just have the user pass in a env var to turn it on if they want.

let me know if we want to change the default to false and have it be enabled by a system property and I can do that. That is simple to do. I just figured 'we are a system that collects metrics - why would we not want to enable metrics by default?

jshaughn:
This was sufficient: "By default WildFly does not enable statistics and in this state every metric that Prometheus exporter will collect and scrape for these subsystems will always be 0".

PR #426, opened 2017-12-12
Update docker agent to work in domain mode

josejulio:
Also updates `start-with-hawkular.sh` to expect server url instead of hostname.
This allows to specify https (if needed) or a different port.


jmazzitelli:
LGTM

PR #427, opened 2017-12-15
HAWKULAR-1349 Let JmxScrape fetch individual MBeans

lucasponce:
Context for this fix is the following:
- Wildfly/EAP has an MBean which is special and may not appear in queryNames/queryMBeans.
- This MBean is important for availability information.
Instead to "hack" and add any kind of "hardcoded" name in the code, I think a better approach is to let in the config to query patterns names (with wildcards) and individual instances.
Then, this can be prepared for future use cases, too.
Note, this will be fixed in Wildfly Core, but we need to live with WF10, WF11 and EAP6/EAP7 which will have this (and other similar) situations.

Also, this fix needs another PR on commons to modify the jmx-exporter configuration.

### DIFF
--- a/hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/prometheus/JmxCollector.java
+++ b/hawkular-agent-core/src/main/java/org/hawkular/agent/monitor/prometheus/JmxCollector.java
@@ -1,0 +1,492 @@
+/*
+ * Copyright 2015-2017 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


lucasponce:
Note 2: I'm going also to propose the fix on prometheus repo, but as that process can be long and would need global release, let's use this clone.
Once prometheus fixes this, I will change this again.

jshaughn:
@lucasponce  The code changes look good.  I'd like only to see a few more breadcrumbs in the forked classes.  In class comments maybe add link to jira issue and also link to the exact P github source that we forked (V2.0 version, I presume).  And maybe inline a short comment indicating the change-points.  This could help us later understanding why we did this and where the actual changes are from the base code.

lucasponce:
@jshaughn Updated the comments with the proper documentation for this change.

jshaughn:
LGTM, thanks for the updates.

